PR #2, opened 2024-04-29
feat(codecv2): add zstd compression and increase chunks num per batch post compression

colinlyguo:

add `codecv2`.

### PR title

- [x] feat: A new feature

### Breaking change label

### DIFF
--- a/encoding/da.go
+++ b/encoding/da.go
@@ -91,6 +91,8 @@ func ConvertTxDataToRLPEncoding(txData *types.TransactionData, useMockTxData boo
 		return nil, fmt.Errorf("failed to decode txData.Data: data=%v, err=%w", txData.Data, err)
 	}
 
+	// This mock param is only used in testing comparing batch challenges with standard test cases.
+	// These tests use this param to set the tx data for convenience.
 	if useMockTxData {
 		return data, nil
 	}


encoding/codecv2/codecv2.go
Thegaram, line 1: Any way we could share some of the repeated code? (If no easy way then duplication is also acceptable.)

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+	dataHash := crypto.Keccak256Hash(dataBytes)
+	return dataHash, nil
+}
+
+// constructBlobPayload constructs the 4844 blob payload.
+func constructBlobPayload(chunks []*encoding.Chunk) (*kzg4844.Blob, common.Hash, *kzg4844.Point, error) {
+	// metadata consists of num_chunks (2 bytes) and chunki_size (4 bytes per chunk)
+	metadataLength := 2 + MaxNumChunks*4
+
+	// the raw (un-padded) blob payload
```
Thegaram, line 307:
```suggestion
	// the raw (un-compressed and un-padded) blob payload
```

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+	}
+
+	dataHash := crypto.Keccak256Hash(dataBytes)
+	return dataHash, nil
+}
+
+// constructBlobPayload constructs the 4844 blob payload.
+func constructBlobPayload(chunks []*encoding.Chunk) (*kzg4844.Blob, common.Hash, *kzg4844.Point, error) {
+	// metadata consists of num_chunks (2 bytes) and chunki_size (4 bytes per chunk)
+	metadataLength := 2 + MaxNumChunks*4
```
Thegaram, line 305: So this changed from 62 (which when padded nicely aligns to 64) to 182, any concern about this? (I guess not, it didn't need to be aligned, and now with compression it's even less relevant, just double checking.)

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+
+	// the raw (un-padded) blob payload
+	blobBytes := make([]byte, metadataLength)
```
Thegaram, line 308: Did you plan to rename this to `batchBytes`?

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+	}
+	return batchBytes, nil
+}
+
+func compressScrollBatchBytes(batchBytes []byte) ([]byte, error) {
+	srcSize := C.uint64_t(len(batchBytes))
+	outbuf := make([]byte, len(batchBytes))
+	outbufSize := C.uint64_t(len(batchBytes))
+
+	err := C.compress_scroll_batch_bytes(
```
Thegaram, line 598: I guess we don't need to `Free` the returned `err` string, since it reuses `outbuf` which will be garbage collected by Go, right?

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+		if chunkSize := len(batchBytes) - currentChunkStartIndex; chunkSize != 0 {
+			binary.BigEndian.PutUint32(batchBytes[2+4*chunkID:], uint32(chunkSize))
+		}
+	}
+	return batchBytes, nil
+}
+
+func compressScrollBatchBytes(batchBytes []byte) ([]byte, error) {
+	srcSize := C.uint64_t(len(batchBytes))
+	outbuf := make([]byte, len(batchBytes))
```
Thegaram, line 595: Is the output guaranteed to be shorter (or equal) length as the input?

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,622 @@
...
+	return calculatePaddedBlobSize(uint64(len(blobBytes))), nil
+}
+
+// EstimateBatchL1CommitBlobSize estimates the total size of the L1 commit blob for a batch.
+func EstimateBatchL1CommitBlobSize(b *encoding.Batch) (uint64, error) {
+	batchBytes, err := constructBatchPayload(b.Chunks)
+	if err != nil {
+		return 0, err
+	}
+	blobBytes, err := compressScrollBatchBytes(batchBytes)
```
Thegaram, line 519: Will it be expensive to call this many times repeatedly?

encoding/codecv2/codecv2_test.go
```diff
@@ -0,0 +1,806 @@
...
+	"github.com/scroll-tech/go-ethereum/common"
+	"github.com/scroll-tech/go-ethereum/core/types"
+	"github.com/scroll-tech/go-ethereum/crypto/kzg4844"
+	"github.com/stretchr/testify/assert"
+
+	"github.com/scroll-tech/da-codec/encoding"
+	"github.com/scroll-tech/da-codec/encoding/codecv0"
+)
+
+func TestCodecV1BlockEncode(t *testing.T) {
```
Thegaram, line 19:
```suggestion
func TestCodecV2BlockEncode(t *testing.T) {
```

Thegaram (dismissed):

encoding/codecv1/codecv1.go
```diff
@@ -524,3 +522,29 @@ func chunkL1CommitBlobDataSize(c *encoding.Chunk) (uint64, error) {
...
+}
+
+// calculatePaddedBlobSize calculates the required size on blob storage
+// where every 32 bytes can store only 31 bytes of actual data, with the first byte being zero.
+func calculatePaddedBlobSize(dataSize uint64) uint64 {
+	paddedSize := (dataSize / 31) * 32
+
+	if dataSize%31 != 0 {
+		paddedSize += 1 + dataSize%31 // Add 1 byte for the first empty byte plus the remainder bytes
+	}
```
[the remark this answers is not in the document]
colinlyguo: Yes. because this field is also used for blob usage analysis, thus making it more accurate.

Dockerfile
```diff
@@ -0,0 +1,38 @@
+ARG GO_VERSION=1.21
+ARG RUST_VERSION=nightly-2023-12-03
+ARG CARGO_CHEF_TAG=0.1.41
+
+FROM ubuntu:20.04
+
+RUN apt-get update && ln -fs /usr/share/zoneinfo/America/New_York /etc/localtime
```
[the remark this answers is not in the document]
colinlyguo: I reuse the Dockerfile here: https://github.com/scroll-tech/go-ethereum/blob/scroll-v5.3.11/Dockerfile

It ensures that no matter where the image runs, it always uses `/usr/share/zoneinfo/America/New_York` time. But this Dockerfile is only used for local testing, thus I'll remove the timezone setting.

encoding/codecv2/codecv2.go
colinlyguo: agree. and also cc @roynalnaruto

rs/libscroll_zstd.so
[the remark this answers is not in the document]
HAOYUatHZ: we should not commit these binaries to git

encoding/codecv2/codecv2.go
colinlyguo: I found the naming is used in other places, like:
```
batchBytes := daBatch.Encode()
```
so keep it as `blobBytes` here. I think `blobBytes`, `uncompressedBlobBytes` or `rawBlobBytes` will not cause confusion too, thus no need to change the naming.

encoding/codecv2/codecv2_test.go
```diff
@@ -0,0 +1,806 @@
...
+
+		blockv1, err := NewDABlock(trace, 0)
+		assert.NoError(t, err)
+		encodedv1 := hex.EncodeToString(blockv1.Encode())
+
+		assert.Equal(t, encodedv0, encodedv1)
+	}
+}
+
+func TestCodecV1ChunkEncode(t *testing.T) {
```
colinlyguo, line 74:
```suggestion
func TestCodecV2ChunkEncode(t *testing.T) {
```

encoding/codecv2/codecv2_test.go
```diff
@@ -0,0 +1,806 @@
...
+
+	trace = readBlockFromJSON(t, "../testdata/blockTrace_07.json")
+	originalChunk = &encoding.Chunk{Blocks: []*encoding.Block{trace}}
+	chunk, err = NewDAChunk(originalChunk, 0)
+	assert.NoError(t, err)
+	encoded = hex.EncodeToString(chunk.Encode())
+	assert.Equal(t, "01000000000000001100000000646b6ed0000000000000000000000000000000000000000000000000000000000000000000000000007a120001010101", encoded)
+}
+
+func TestCodecV1ChunkHash(t *testing.T) {
```
colinlyguo, line 129:
```suggestion
func TestCodecV2ChunkHash(t *testing.T) {
```

encoding/codecv2/codecv2_test.go
```diff
@@ -0,0 +1,806 @@
...
+
+		_, y, err := kzg4844.ComputeProof(b, *z)
+		assert.NoError(t, err)
+		actualY := hex.EncodeToString(y[:])
+		assert.Equal(t, tc.expectedy, actualY)
+
+	}
+}
+
+func TestCodecV1BatchBlobDataProof(t *testing.T) {
```
colinlyguo, line 595:
```suggestion
func TestCodecV2BatchBlobDataProof(t *testing.T) {
```

encoding/codecv2/codecv2_test.go
```diff
@@ -0,0 +1,806 @@
...
+
+	originalBatch.TotalL1MessagePoppedBefore = 10
+	batch, err = NewDABatch(originalBatch)
+	assert.NoError(t, err)
+	assert.Equal(t, "0000000000000000000000000000000000000000000000000000000007fffffe", hex.EncodeToString(batch.SkippedL1MessageBitmap))
+	assert.Equal(t, 32, int(batch.L1MessagePopped))
+	assert.Equal(t, 42, int(batch.TotalL1MessagePopped))
+}
+
+func TestCodecV1ChunkAndBatchBlobSizeEstimation(t *testing.T) {
```
colinlyguo, line 754:
```suggestion
func TestCodecV2ChunkAndBatchBlobSizeEstimation(t *testing.T) {
```

rs/Cargo.lock
```diff
@@ -0,0 +1,4420 @@
...
+checksum = "18cf6c6447f813ef19eb450e985bcce6705f9ce7660db221b59093d15c79c4b7"
+dependencies = [
+ "proc-macro-crate 1.3.1",
+ "proc-macro2",
+ "quote",
+ "syn 1.0.109",
+]
+
+[[package]]
+name = "scroll_zstd"
```
colinlyguo, line 3088:
```suggestion
name = "scroll-zstd"
```

Dockerfile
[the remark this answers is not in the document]
colinlyguo: removed in [6aefc05](https://github.com/scroll-tech/da-codec/pull/2/commits/6aefc0506609c08da6215f4e2d5b017702ed90f6).

encoding/codecv2/codecv2.go
colinlyguo: nice catch, it's not mentioned in the doc. will reserve 256 more bytes for output.

Dockerfile
```diff
@@ -0,0 +1,36 @@
...
+ARG RUST_VERSION
+RUN rustup toolchain install ${RUST_VERSION}
+ARG CARGO_CHEF_TAG
+RUN cargo install cargo-chef --locked --version ${CARGO_CHEF_TAG} \
+    && rm -rf $CARGO_HOME/registry/
+
+# Install Go
+ARG GO_VERSION
+RUN rm -rf /usr/local/go
+RUN wget https://go.dev/dl/go${GO_VERSION}.1.linux-amd64.tar.gz
```
georgehao, line 33:
```suggestion
RUN wget https://go.dev/dl/go${GO_VERSION}.linux-amd64.tar.gz
```

Dockerfile
```diff
@@ -0,0 +1,36 @@
+ARG GO_VERSION=1.21
```
georgehao, line 1:
```suggestion
ARG GO_VERSION=1.21.1
```

encoding/codecv2/codecv2.go
georgehao: Maybe can consider this https://github.com/scroll-tech/scroll/pull/1178#discussion_r1522956248

encoding/codecv2/codecv2.go
colinlyguo: added in [e5d513b](https://github.com/scroll-tech/da-codec/pull/2/commits/e5d513b0da25c235182819e5ec3e2b205cb11757)

rs/rs_zstd.h
```diff
@@ -0,0 +1,4 @@
+#include <stdint.h>
+
+char* compress_scroll_batch_bytes(uint8_t* src, uint64_t src_size,
```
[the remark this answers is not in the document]
colinlyguo: refactored the logic in: [e5d513b](https://github.com/scroll-tech/da-codec/pull/2/commits/e5d513b0da25c235182819e5ec3e2b205cb11757)

rs/src/lib.rs
```diff
@@ -0,0 +1,50 @@
...
+    let mut encoder = init_zstd_encoder();
+    encoder
+        .set_pledged_src_size(Some(src.len() as u64))
+        .expect("infallible");
+
+    let ret = encoder.write_all(src);
+    let ret = ret.and_then(|_|encoder.finish());
+    if let Err(e) = ret {
+        out_as_err(e.to_string().as_str(), out);
+        return output_buf as *const c_char
```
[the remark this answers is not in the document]
colinlyguo: I think it's used for memory allocation. rust side does not manage memory, only leaving this task to golang.
so for happy path, `output_buf` is used to store compressed string, for unhappy path, `output_buf` is used to store error message.

encoding/codecv2/codecv2.go
colinlyguo: Yes. I think it's the original motivation for implementing the `out_as_err` function (reusing `outbuf`). The code snippet is from @noel2004.

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN rustup toolchain install ${RUST_VERSION}
+ARG CARGO_CHEF_TAG
+RUN cargo install cargo-chef --locked --version ${CARGO_CHEF_TAG} \
+    && rm -rf $CARGO_HOME/registry/
+
+# Install Go
+ARG GO_VERSION
+RUN rm -rf /usr/local/go
+RUN wget https://go.dev/dl/go${GO_VERSION}.1.linux-amd64.tar.gz
+RUN tar -C /usr/local -xzf go${GO_VERSION}.1.linux-amd64.tar.gz
```
colinlyguo, line 36:
```suggestion
RUN tar -C /usr/local -xzf go${GO_VERSION}.linux-amd64.tar.gz
```

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+RUN rm go${GO_VERSION}.1.linux-amd64.tar.gz
```
colinlyguo, line 37:
```suggestion
RUN rm go${GO_VERSION}.linux-amd64.tar.gz
```

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
+package codecv2
+
+/*
+#cgo LDFLAGS: -L../../rs -lm -ldl -lscroll_zstd
+#include <stdlib.h>
+#include "../../rs/rs_zstd.h"
```
georgehao, line 6: it'll be better that generate the `rs_zstd.h` and `libzktrie.so` to `encoding` folder.

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
...
+	BLSModulus *big.Int
+
+	// BlobDataProofArgs defines the argument types for `_blobDataProof` in `finalizeBatchWithProof4844`.
+	BlobDataProofArgs abi.Arguments
+
+	// MaxNumChunks is the maximum number of chunks that a batch can contain.
+	MaxNumChunks int = 45
+)
+
+func init() {
```
georgehao, line 42: change `init` to `Newxxx() error`, outside can get the error.

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
...
+	// initialize modulus
+	modulus, success := new(big.Int).SetString("52435875175126190479447740508185965837690552500527637822603658699938581184513", 10)
```
georgehao, line 44: `52435875175126190479447740508185965837690552500527637822603658699938581184513` represents what?

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
...
+		binary.BigEndian.PutUint64(bytes[40:], b.BaseFee.Uint64())
+	}
+	binary.BigEndian.PutUint64(bytes[48:], b.GasLimit)
+	binary.BigEndian.PutUint16(bytes[56:], b.NumTransactions)
+	binary.BigEndian.PutUint16(bytes[58:], b.NumL1Messages)
+	return bytes
+}
+
+// DecodeDABlock takes a byte slice and decodes it into a DABlock.
+func DecodeDABlock(bytes []byte) (*DABlock, error) {
```
georgehao, line 147: why `DecodeDABlock` don't have the `* DABlock` receiver?

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
...
+		GasLimit:        binary.BigEndian.Uint64(bytes[48:56]),
+		NumTransactions: binary.BigEndian.Uint16(bytes[56:58]),
+		NumL1Messages:   binary.BigEndian.Uint16(bytes[58:60]),
+	}
+
+	return block, nil
+}
+
+// NewDAChunk creates a new DAChunk from the given encoding.Chunk and the total number of L1 messages popped before.
+func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DAChunk, error) {
```
georgehao, line 165: split contents:
*  `DABlock` to `dablock.go`
*  `DAChunk` to `dachunk.go`

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,618 @@
...
+				hashBytes, err := hex.DecodeString(txHash)
+				if err != nil {
+					return common.Hash{}, err
+				}
+				if len(hashBytes) != 32 {
+					return common.Hash{}, fmt.Errorf("unexpected hash: %s", txData.TxHash)
+				}
+				dataBytes = append(dataBytes, hashBytes...)
+			}
+		}
```
georgehao, line 225:
```suggestion
		for _, txData := range blockTxs {
		        if txData.Type != types.L1MessageTxType {
		             continue
		        }
		        
			txHash := strings.TrimPrefix(txData.TxHash, "0x")
			hashBytes, err := hex.DecodeString(txHash)
			if err != nil {
				return common.Hash{}, err
			}
			if len(hashBytes) != 32 {
				return common.Hash{}, fmt.Errorf("unexpected hash: %s", txData.TxHash)
			}
			dataBytes = append(dataBytes, hashBytes...)
		}
```

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,507 @@
...
+	}
+
+	return &daBatch, nil
+}
+
+// ComputeBatchDataHash computes the data hash of the batch.
+// Note: The batch hash and batch data hash are two different hashes,
+// the former is used for identifying a badge in the contracts,
+// the latter is used in the public input to the provers.
+func ComputeBatchDataHash(chunks []*encoding.Chunk, totalL1MessagePoppedBefore uint64) (common.Hash, error) {
```
georgehao, line 250: surround with `receiver` will be better

encoding/codecv2/codecv2.go
colinlyguo: refactored in [e9dbcf9](https://github.com/scroll-tech/da-codec/pull/2/commits/e9dbcf9b00abb7ee6c6639a11caec35dbc226e59)

further increasing code reuse needs merging `DABlock`, `DAChunk` and `DABatch` in `codecv0`, `codecv1`, `codecv2`, which would change the functions provided. can be considered and discussed later.

@georgehao's comment is also a promising refactor, using interfaces and reflection, this can be discussed together with the above point (merging some structure) in a future PR. because they are all big changes.

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,507 @@
...
+	return codecv1.ComputeBatchDataHash(chunks, totalL1MessagePoppedBefore)
```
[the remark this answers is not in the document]
colinlyguo: based on this: https://github.com/scroll-tech/da-codec/pull/2#discussion_r1590208866
only `codecv2` would reuse `codecv1`, because they only differ in `version` and `compression mode`.

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,507 @@
...
+}
+
+// constructBlobPayload constructs the 4844 blob payload.
+func constructBlobPayload(chunks []*encoding.Chunk) (*kzg4844.Blob, common.Hash, *kzg4844.Point, error) {
```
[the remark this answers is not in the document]
colinlyguo: I think it's OK not to add a receiver to pure computing functions, because it can be used easier without initializing a receiver, and also it can reuse codecv1's implementations.

encoding/codecv2/codecv2.go
colinlyguo: The only place that uses it is l2geth rollup-sync-service:
```
daBlock, err := codecv0.DecodeDABlock(chunk[startIdx:endIdx])
if err != nil {
	return nil, err
}
```
adding a receiver is also OK. the difference is that the invoke would change to:
```
daBlock := &codecv0.DABlock{}
err := daBlock.DecodeDABlock(chunk[startIdx:endIdx])
if err != nil {
	return nil, err
}
```
so I think we can keep the function as it is.

encoding/codecv2/codecv2.go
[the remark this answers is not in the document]
georgehao: if you prefer this function to a common method, maybe put it into `utils` or `common`

encoding/codecv2/codecv2.go
colinlyguo: it's a modulo used by blob. search BLS_MODULUS in: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4844.md

encoding/codecv2/codecv2.go
georgehao: I prefer to keep consistency, because `da-codec` is a public pkg.

rs/src/lib.rs
```diff
@@ -0,0 +1,51 @@
...
+    src: *const c_uchar,
+    src_size: u64,
+    output_buf: *mut c_uchar,
+    output_buf_size: *mut u64,
+) -> *const c_char {
+    let buf_size = *output_buf_size;
+    let src = unsafe { slice::from_raw_parts(src, src_size as usize) };
+    let out = unsafe { slice::from_raw_parts_mut(output_buf, buf_size as usize)};
+
+    let mut encoder = init_zstd_encoder();
```
georgehao, line 32: is it possible `init_zstd_encoder` once?

rs/.gitignore
```diff
@@ -0,0 +1,2 @@
+/target
```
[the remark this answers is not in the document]
colinlyguo: renamed to `libzstd` in [c2a0852](https://github.com/scroll-tech/da-codec/pull/2/commits/c2a085213dfa5eb66393cd2f9f7cf548c5f1cca3).

encoding/codecv2/codecv2.go
```diff
@@ -0,0 +1,507 @@
...
+		for _, block := range chunk.Blocks {
+			for _, tx := range block.Transactions {
+				if tx.Type != types.L1MessageTxType {
+					// encode L2 txs into blob payload
+					rlpTxData, err := encoding.ConvertTxDataToRLPEncoding(tx)
+					if err != nil {
+						return nil, common.Hash{}, nil, err
+					}
+					blobBytes = append(blobBytes, rlpTxData...)
+				}
```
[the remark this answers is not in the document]
colinlyguo: refactored in [67cc822](https://github.com/scroll-tech/da-codec/pull/2/commits/67cc822e9db69b8d112f610b65167e3501729ba3).

encoding/codecv2/codecv2.go
colinlyguo: refactored in [67cc822](https://github.com/scroll-tech/da-codec/pull/2/commits/67cc822e9db69b8d112f610b65167e3501729ba3).

encoding/codecv2/codecv2.go
colinlyguo: changed in [d6ff1c1](https://github.com/scroll-tech/da-codec/pull/2/commits/d6ff1c12eca2760e8e909c1a26636b464ec69979).

libzstd/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+edition = "2021"
+
+[lib]
+crate-type = ["cdylib"]
+
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+aggregator = { git = "https://github.com/scroll-tech/zkevm-circuits.git", branch = "feat/rs_zstd_export" }
```
colinlyguo, line 13: is the encoder thread-safe (ask this because we only want to initialize an encoder globally once, not know if needs a lock), and would we use a stable branch later? currently, it's a feature branch.
@noel2004

rs/src/lib.rs
colinlyguo: nice catch. I tried with once_cell but failed. @noel2004 have we supported single initialization of the encoder? It's critical for performance.

encoding/codecv2/codecv2.go
colinlyguo: Makes sense. we can add another PR to refactor this because it would become harder to review in the same PR.

rs/src/lib.rs
noel2004: I think you can not simply reuse the `zstd_encoder`. According to the the underlying c code, this woould initialize a new context object and should be only used to compress single input bytes.

Have this really caused some impaction to the performance?

libzstd/Cargo.toml
noel2004: the encoder object itself is not thread-safe (and it should be not possible to handle input concurrencily).

The `da-compression` is not merged into stable branch yet so we can only use a feature branch now.

encoding/codecv2/codecv2.go
noel2004: Yes. I just use the `outbuf` for outputing the error string (if there is any)

rs/src/lib.rs
georgehao: it's ok if can't reuse it. just look here superficially, and think it will be better to reuse it.

libzstd/Cargo.toml
colinlyguo: got it.

rs/src/lib.rs
colinlyguo: Yes. @noel2004  From the doc:
```
  When compressing many times,
  it is recommended to allocate a context just once,
  and re-use it for each successive compression operation.
  This will make workload friendlier for system's memory.
  Note : re-using context is just a speed / resource optimization.
         It doesn't change the compression ratio, which remains identical.
  Note 2 : In multi-threaded environments,
         use one different context per thread for parallel execution.
```
Initializing the context only once for each thread is recommended. Especially for Golang's memory management:
```
This will make workload friendlier for system's memory.
```

encoding/codecv2/codecv2.go
Thegaram: Why 256 specifically?

colinlyguo: It's just a heuristic value set blindly (considering memory management, performance, normal error message length, and zstd header length to prevent output overflow (have not checked the doc yet)). can refine it.

encoding/codecv2/codecv2.go
colinlyguo: got it. so I can resolve this comment, right?

encoding/codecv2/codecv2.go
colinlyguo: `init()` and `log.Crit()` are removed in [1f826c0](https://github.com/scroll-tech/da-codec/pull/2/commits/1f826c00dd47706c55a4cf6b580c01a3bc05f671).

georgehao (dismissed):

rs/libscroll_zstd.so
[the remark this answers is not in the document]
colinlyguo: they are built step by step following this `Readme.md`: https://github.com/scroll-tech/da-codec/blob/1f826c00dd47706c55a4cf6b580c01a3bc05f671/README.md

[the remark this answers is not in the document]
colinlyguo: based on ubuntu-20.04.

georgehao (approved):

HAOYUatHZ (approved):

georgehao (approved):

Dockerfile
HAOYUatHZ, line 1: this only installs go & rust? what's the purpose of this dockerfile?

colinlyguo: It's used for running the local unit test, see the readme (`make run`): https://github.com/scroll-tech/da-codec/blob/f8b88ce4e7f047624756e75ccb46949eda0caa70/README.md
I also use this container to build .so files, see this pre-release: https://github.com/scroll-tech/da-codec/releases/tag/v0.0.0-rc0-ubuntu20.04

colinlyguo: removed by using `scrolltech/go-rust-builder:go-1.21-rust-nightly-2023-12-03`.

encoding/da.go
```diff
@@ -19,15 +19,31 @@ const (
 	// CodecV1 represents the version 1 of the encoder and decoder.
 	CodecV1
 
+	// CodecV2 represents the version 2 of the encoder and decoder.
+	CodecV2
+
 	// txTypeTest is a special transaction type used in unit tests.
 	txTypeTest = 0xff
```
Thegaram, line 26: It looks weird to have `txTypeTest` in the same "enum declaration" as the codec versions, can we separate these?

run_test.sh
Thegaram, line 1: What's the purpose of this script?

Thegaram, line 1: What's the purpose of this script?

encoding/codecv2/codecv2.go
Thegaram: No idea in mind, it seems impossible to do any incremental estimation when we use compression. If this is not a concern in your tests, then we don't need to worry about it. But maybe we can add some metrics to track the runtime in the monorepo, just in case.

HAOYUatHZ (dismissed):
TODO: add github auto-release action in the future

run_test.sh
colinlyguo: to run unit tests in CI. I tried putting all cmds in the Makefile, but each cmd is run within an independent shell env, and the env setting will not take effect in later cmds. so adding a script for convenience.

encoding/da.go
colinlyguo: Yeah. it's a hack way, only used in testing the challenge points with some "assigned payload", better remove it.

One way is to add another `constructBlobPayloadWithAssignedTxData`, adding another ConvertTxDataToRLPEncoding to assign the input directly as the rlp-encoding.

Or some smaller changes to achieve this?

encoding/codecv2/codecv2.go
colinlyguo: Yes. will add some in the monorepo.

rs/src/lib.rs
colinlyguo: closed this comment, but it's still a nice-to-have feature.

Thegaram (approved):

HAOYUatHZ (approved):

PR #4, opened 2024-05-26
feat: induce zstd

noel2004:
Now we have put the encoding parameters and the dependency of rs-zstd (our forking) into the `libzstd` crate.

So we could be able to build a very light weight .so now


libzstd/encoder/src/lib.rs
```diff
@@ -0,0 +1,46 @@
+use zstd::stream::Encoder;
```
colinlyguo, line 1: could we also use the param setting "inside" `.so` file, because when upgrading `zstd`, also need to sync params in this file in the meantime.

libzstd/encoder/src/lib.rs
```diff
@@ -0,0 +1,46 @@
...
+#[allow(dead_code)]
+pub const CL_WINDOW_LIMIT: usize = 17;
+
+/// zstd block size target.
+pub const N_BLOCK_SIZE_TARGET: u32 = 124 * 1024;
+
+/// Maximum number of blocks that we can expect in the encoded data.
+pub const N_MAX_BLOCKS: u64 = 10;
+
+/// Zstd encoder configuration
```
georgehao, line 18: a quick question: the `encoder` is another crate? it isn't a part of libstd?

noel2004: since `aggregator` has to depend on this crate. It is not suitable to make it a dynamic lib.

So we provide `encoder` for `aggregator`, then wrap `encoder` (a static lib) into libzstd (a dynamic lib for Golang side)

georgehao: got it

georgehao (approved):

colinlyguo (approved):

PR #9, opened 2024-06-07
build: update l2geth dependency

colinlyguo:

Update zktrie version to a stable version, preventing `go mod tidy` from reverting zktrie version in l2geth repo.
Since l2geth repo uses go 1.21, updating the dependency requires updating go dependency to 1.21, too.

### PR title

- [x] build: Changes that affect the build system or external dependencies (example scopes: yarn, eslint, typescript)

### Breaking change label


go.mod
```diff
@@ -1,14 +1,13 @@
 module github.com/scroll-tech/da-codec
 
-go 1.20
+go 1.21
 
 require (
-	github.com/scroll-tech/go-ethereum v1.10.14-0.20240426041101-a860446ebaea
+	github.com/scroll-tech/go-ethereum v1.10.14-0.20240607130425-e2becce6a1a4
```
omerfirmak, line 6: are we able to use tags here? `scroll-v5.3.41`

Thegaram (approved):

go.mod
colinlyguo: tried with v5.3.41. the tag is incompatible with golang's format. thus changing it to use commit hash.

colinlyguo:
```
go: errors parsing go.mod:
go.mod:6:2: require github.com/scroll-tech/go-ethereum: version "v5.3.41" invalid: should be v0 or v1, not v5
```

georgehao (approved):

omerfirmak (approved):

PR #10, opened 2024-06-11
feat: add decoding methods

NazariiDenha:

*Describe your change. Make sure to answer these three questions: What does this PR do? Why does it do it? How does it do it?*

Implement some decoding method, that will be used by `l2geth` during the process of syncing from DA

### PR title

- [ ] build: Changes that affect the build system or external dependencies (example scopes: yarn, eslint, typescript)
- [ ] ci: Changes to our CI configuration files and scripts (example scopes: vercel, github, cypress)
- [ ] docs: Documentation-only changes
- [x] feat: A new feature
- [ ] fix: A bug fix
- [ ] perf: A code change that improves performance
- [ ] refactor: A code change that doesn't fix a bug, or add a feature, or improves performance
- [ ] style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
- [ ] test: Adding missing tests or correcting existing tests

### Breaking change label

- [ ] Yes


NazariiDenha:
I will also add some tests later

NazariiDenha:
I will ass decoding for codecv2 and then will be ready for review

NazariiDenha:
ready for review

encoding/codecv2/codecv2.go
```diff
@@ -341,6 +385,46 @@ func constructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
...
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	compressedBytes := codecv1.BytesFromBlobCanonical(blob)
+	magics := []byte{0x28, 0xb5, 0x2f, 0xfd}
+
+	blobBytes, err := decompressScrollBatchBytes(append(magics, compressedBytes[:]...))
+	if err != nil {
+		return err
+	}
+
+	numChunks := int(binary.BigEndian.Uint16(blobBytes[0:2]))
```
colinlyguo, line 398: could we reuse codecv1's implementation by some refactor? seems like MaxNumChunks is 45 in codecv2, and 15 in codecv1.

I'm good at both styles, if there is no neat refactor.

encoding/codecv2/codecv2.go
```diff
@@ -523,3 +607,30 @@ func compressScrollBatchBytes(batchBytes []byte) ([]byte, error) {
...
+
+	r := bytes.NewReader(compressedBytes)
+	zr, err := zstd.NewReader(r)
+	if err != nil {
+		return nil, err
+	}
+	defer zr.Close()
+
+	for {
+		i, _ := zr.Read(batchOfBytes) //
```
colinlyguo, line 626: error handling for completeness sake.

encoding/codecv2/codecv2.go
```diff
@@ -523,3 +607,30 @@ func compressScrollBatchBytes(batchBytes []byte) ([]byte, error) {
...
+	batchOfBytes := make([]byte, readBatchSize)
+
+	r := bytes.NewReader(compressedBytes)
+	zr, err := zstd.NewReader(r)
+	if err != nil {
+		return nil, err
+	}
+	defer zr.Close()
+
+	for {
```
colinlyguo, line 625: maybe the for loop is not needed here? if the size of the blob remains <= 131072.

encoding/codecv2/codecv2.go
NazariiDenha: difference not only in NumChunks, codecv2 also uses decompression

encoding/codecv2/codecv2.go
NazariiDenha: the size of blob remains <= 131072, but size of decompressed data maybe 131072*3? if you can provide strict upperbound for the size of decompressed data (e.g. data before compressing) then I can change it and remove cycle

colinlyguo: got it.

encoding/codecv2/codecv2.go
```diff
@@ -410,3 +492,30 @@ func CalculatePaddedBlobSize(dataSize uint64) uint64 {
...
+		i, err := zr.Read(batchOfBytes)
+		res = append(res, batchOfBytes[:i]...) // append already decoded bytes even if we meet error
+		if i < readBatchSize || err != nil {
```
colinlyguo, line 513: What about returning the error `if err != nil`? otherwise, the error is not handled.

encoding/codecv0/codecv0.go
```diff
@@ -179,6 +187,63 @@ func (c *DAChunk) Encode() ([]byte, error) {
 	return chunkBytes, nil
 }
 
+// DecodeDAChunksRawTx takes a byte slice and decodes it into a []DAChunkRawTx.
```
colinlyguo, line 190:
```suggestion
// DecodeDAChunksRawTx takes a byte slice and decodes it into a []*DAChunkRawTx.
```

encoding/codecv0/codecv0.go
```diff
@@ -179,6 +187,63 @@ func (c *DAChunk) Encode() ([]byte, error) {
...
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		var transactions []types.Transactions
+		currentIndex := 1 + numBlocks*BlockContextByteSize
+		for _, block := range blocks {
+			var blockTransactions types.Transactions
+			var txNum int = int(block.NumTransactions - block.NumL1Messages)
```
colinlyguo, line 218:
```suggestion
			txNum := int(block.NumTransactions - block.NumL1Messages)
```

encoding/codecv3/codecv3.go
```diff
@@ -52,6 +57,11 @@ func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DACh
 	return codecv2.NewDAChunk(chunk, totalL1MessagePoppedBefore)
 }
 
+// DecodeDAChunksRawTx takes a byte slice and decodes it into a []DAChunkRawTx.
+func DecodeDAChunksRawTx(bytes [][]byte) ([]*codecv2.DAChunkRawTx, error) {
```
colinlyguo, line 61:
```suggestion
// DecodeDAChunksRawTx takes a byte slice and decodes it into a []*DAChunkRawTx.
func DecodeDAChunksRawTx(bytes [][]byte) ([]*DAChunkRawTx, error) {
```

encoding/codecv3/codecv3.go
```diff
@@ -122,6 +132,11 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 	return codecv2.ConstructBlobPayload(chunks, useMockTxData)
 }
 
+// DecodeTxsFromBlob decodes txs from blob bytes and writes to chunks
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*codecv2.DAChunkRawTx) error {
```
colinlyguo, line 136:
```suggestion
func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
```

encoding/codecv3/codecv3.go
colinlyguo: or you can just remove `DecodeDAChunksRawTx` and `DecodeTxsFromBlob`, because by using `type DAChunkRawTx codecv2.DAChunkRawTx`, these functions can be reused.

encoding/codecv2/codecv2.go
```diff
@@ -64,6 +72,40 @@ func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DACh
 	return codecv1.NewDAChunk(chunk, totalL1MessagePoppedBefore)
 }
 
+// DecodeDAChunksRawTx takes a byte slice and decodes it into a []DAChunkRawTx.
```
colinlyguo, line 75:
```suggestion
// DecodeDAChunksRawTx takes a byte slice and decodes it into a []*DAChunkRawTx.
```

encoding/codecv1/codecv1.go
```diff
@@ -98,6 +103,40 @@ func (c *DAChunk) Encode() []byte {
 	return chunkBytes
 }
 
+// DecodeDAChunksRawTx takes a byte slice and decodes it into a []DAChunkRawTx.
```
colinlyguo, line 106:
```suggestion
// DecodeDAChunksRawTx takes a byte slice and decodes it into a []*DAChunkRawTx.
```

encoding/codecv1/codecv1.go
```diff
@@ -98,6 +103,40 @@ func (c *DAChunk) Encode() []byte {
...
+				return nil, err
+			}
+		}
+
+		var transactions []types.Transactions
+
+		chunks = append(chunks, &DAChunkRawTx{
+			Blocks:       blocks,
+			Transactions: transactions,
+		})
```
colinlyguo, line 135:
```suggestion
		chunks = append(chunks, &DAChunkRawTx{
			Blocks:       blocks,
			Transactions: nil, // this field is still empty in the phase of DecodeDAChunksRawTx.
		})
```

colinlyguo: Why separate the decoding into two functions instead of using a function with both `dachunk` and `blob` as input and decoding everything inside?

encoding/codecv1/codecv1.go
```diff
@@ -316,6 +355,104 @@ func MakeBlobCanonical(blobBytes []byte) (*kzg4844.Blob, error) {
...
+	index := 2 + MaxNumChunks*4
+	for chunkID, chunk := range chunks {
+		var transactions []types.Transactions
+		chunkSize := int(binary.BigEndian.Uint32(blobBytes[2+4*chunkID : 2+4*chunkID+4]))
+
+		chunkBytes := blobBytes[index : index+chunkSize]
+		curIndex := 0
+		for _, block := range chunk.Blocks {
+			var blockTransactions types.Transactions
+			var txNum = int(block.NumTransactions - block.NumL1Messages)
```
colinlyguo, line 374:
```suggestion
			txNum := int(block.NumTransactions - block.NumL1Messages)
```

colinlyguo: just a suggestion, and it's better to align these initialization styles in the same repo, another style is found in codecv0's `DecodeDAChunksRawTx`.

encoding/codecv3/codecv3.go
colinlyguo: same as https://github.com/scroll-tech/da-codec/pull/10#discussion_r1690992283, can remove this function.

encoding/codecv2/codecv2.go
NazariiDenha: added a comment explaining this. the error is expected and means that decoder can't decode further compressed data. As explained at the beginning of method "decompress data in stream and in batches of bytes, **because we don't know actual length of compressed data**". So even after error we will have data written in `res` and should return this

encoding/codecv1/codecv1.go
NazariiDenha: I chose to separate it, because blob decoding may differ on different versions and it's not always called together with calldata decoding

colinlyguo: got it.

encoding/codecv2/codecv2.go
colinlyguo: o i c. and yeah it makes sense to add a comment here.

colinlyguo: then what about early return before `res = append(res, batchOfBytes[:i]...)` when EOF?

encoding/codecv2/codecv2.go
```diff
@@ -32,12 +35,17 @@ var BLSModulus = codecv1.BLSModulus
 // MaxNumChunks is the maximum number of chunks that a batch can contain.
 const MaxNumChunks = 45
 
+const BlockContextByteSize = 60
```
jonastheis, line 38:
```suggestion
const BlockContextByteSize = codecv1.BlockContextByteSize
```

encoding/codecv1/codecv1.go
```diff
@@ -26,12 +26,17 @@ var BLSModulus = new(big.Int).SetBytes(common.FromHex("0x73eda753299d7d483339d80
 // MaxNumChunks is the maximum number of chunks that a batch can contain.
 const MaxNumChunks = 15
 
+const BlockContextByteSize = 60
```
jonastheis, line 29:
```suggestion
const BlockContextByteSize = codecv0.BlockContextByteSize
```

encoding/codecv2/codecv2.go
NazariiDenha: `i, err := zr.Read(batchOfBytes)` at this line we can get `err != nil` and `i > 0`. That means that decoder was able to decode `i` bytes and after that catch error. So we still need to append that decoded data and return only after that

NazariiDenha: I added some comments explaining this all

encoding/codecv1/codecv1.go
```diff
@@ -316,6 +355,104 @@ func MakeBlobCanonical(blobBytes []byte) (*kzg4844.Blob, error) {
...
+		chunk.Transactions = transactions
+		index += chunkSize
+	}
+	return nil
+}
+
+var errSmallLength error = fmt.Errorf("length of blob bytes is too small")
+
+// GetNextTx parses blob bytes to find length of payload of next Tx and decode it
+func GetNextTx(bytes []byte, index int) (*types.Transaction, int, error) {
```
roynalnaruto, line 394: Do we test if this can deserialise all the below:
1. EIP-155 legacy transaction
2. EIP-2930 tx
3. EIP-1559 tx

roynalnaruto: I see that we are testing using block traces 02, 03, 04, 05. Do these traces include every tx type that we support?

encoding/codecv2/codecv2.go
```diff
@@ -222,6 +235,46 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
...
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	compressedBytes := codecv1.BytesFromBlobCanonical(blob)
+	magics := []byte{0x28, 0xb5, 0x2f, 0xfd}
+
+	blobBytes, err := decompressScrollBatchBytes(append(magics, compressedBytes[:]...))
+	if err != nil {
+		return err
+	}
+
+	numChunks := int(binary.BigEndian.Uint16(blobBytes[0:2]))
```
roynalnaruto, line 248: I think it will help introducing a new function: `DecodeTxsFromBytes`, where its implementation does not change between v1 and v2.

The `DecodeTxsFromBlob` can internally call its own version's `DecodeTxsFromBytes` after translating the blob to bytes.

In v1 that would be: canonical blob -> blob bytes
In v2 that would be: canonical blob -> blob bytes -> decompress

roynalnaruto: As a result we can then remove the code duplication between v1 and v2 in the `DecodeTxsFromBlob` function. In the future if we detect a bug, it would help updating/reviewing as there would be lesser number of code changes.

roynalnaruto (commented):
Some refactoring suggestions, questions and request for more inline documentation

encoding/codecv1/codecv1.go
NazariiDenha: yes, all tx types included in tests 0 - legacyTx, 2 - DynamicFeeTx, 126 - L1 msg tx

jonastheis (dismissed):

encoding/codecv4/codecv4.go
```diff
@@ -239,6 +248,23 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, enableCompress bool, useMock
...
 }
 
+// DecodeTxsFromBlob decodes txs from blob bytes and writes to chunks
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	rawBytes := encoding.BytesFromBlobCanonical(blob)
+
+	// if first byte is 1 - data compressed, 0 - not compressed
+	if rawBytes[0] == 0x1 {
+		magics := []byte{0x28, 0xb5, 0x2f, 0xfd}
+		blobBytes, err := encoding.DecompressScrollBatchBytes(append(magics, rawBytes[1:]...))
```
roynalnaruto, line 258: The implementation has no issues. I want to point out the naming here though. We decompress a `blob` to get `batch`. So we should have:
```suggestion
		batchBytes, err := encoding.DecompressScrollBatchBytes(append(magics, rawBytes[1:]...))
```

encoding/codecv4/codecv4.go
```diff
@@ -239,6 +248,23 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, enableCompress bool, useMock
...
+		if err != nil {
+			return err
+		}
+		return codecv1.DecodeTxsFromBytes(blobBytes, chunks, MaxNumChunks)
```
roynalnaruto, line 262: And as a result
```suggestion
		return codecv1.DecodeTxsFromBytes(batchBytes, chunks, MaxNumChunks)
```

encoding/da.go
```diff
@@ -354,6 +357,44 @@ func MakeBlobCanonical(blobBytes []byte) (*kzg4844.Blob, error) {
...
+func BytesFromBlobCanonical(blob *kzg4844.Blob) [126976]byte {
+	var blobBytes [126976]byte
+	for from := 0; from < len(blob); from += 32 {
+		copy(blobBytes[from/32*31:], blob[from+1:from+32])
+	}
+	return blobBytes
+}
+
+// DecompressScrollBatchBytes decompresses the given bytes into scroll batch bytes
+func DecompressScrollBatchBytes(compressedBytes []byte) ([]byte, error) {
```
roynalnaruto, line 370: The naming here is a bit confusing. The `blob` bytes are decompressed to get `batch` bytes. But `DecompressScrollBatchBytes` sounds as if we are decompressing the batch bytes. A nit-pick.

encoding/codecv2/codecv2.go
```diff
@@ -217,6 +227,18 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 	return blob, blobVersionedHash, &z, blobBytes, nil
 }
 
+// DecodeTxsFromBlob decodes txs from blob bytes and writes to chunks
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	compressedBytes := encoding.BytesFromBlobCanonical(blob)
+	magics := []byte{0x28, 0xb5, 0x2f, 0xfd}
+
+	blobBytes, err := encoding.DecompressScrollBatchBytes(append(magics, compressedBytes[:]...))
```
roynalnaruto, line 235: Similarly here.
```suggestion
	batchBytes, err := encoding.DecompressScrollBatchBytes(append(magics, compressedBytes[:]...))
```

encoding/codecv2/codecv2.go
```diff
@@ -217,6 +227,18 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
...
+	if err != nil {
+		return err
+	}
+	return codecv1.DecodeTxsFromBytes(blobBytes, chunks, MaxNumChunks)
```
roynalnaruto, line 239:
```suggestion
	return codecv1.DecodeTxsFromBytes(batchBytes, chunks, MaxNumChunks)
```

encoding/codecv1/codecv1.go
```diff
@@ -286,6 +326,99 @@ func constructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
...
+		}
+		chunk.Transactions = transactions
+		index += chunkSize
+	}
+	return nil
+}
+
+// DecodeTxsFromBlob decodes txs from blob bytes and writes to chunks
+func DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	blobBytes := encoding.BytesFromBlobCanonical(blob)
```
roynalnaruto, line 363: I think for v1 this is still OK since `blob` and `batch` were one and the same as of Bernoulli upgrade. It was in Curie upgrade (`v2`) that `blob <- encoded(batch)`

roynalnaruto (commented):
A few minor suggestions and nit-picks.

HAOYUatHZ:
This PR is tested with https://github.com/scroll-tech/go-ethereum/pull/1013, @colinlyguo do you have time to review this PR so that we can merge it?

colinlyguo (approved):

colinlyguo:
> This PR is tested with [scroll-tech/go-ethereum#1013](https://github.com/scroll-tech/go-ethereum/pull/1013), @colinlyguo do you have time to review this PR so that we can merge it?

LGTM. and approved.

colinlyguo:
> Let's get this merged and then integrate into #25. Which means we need to revisit the implementation anyway. Therefore, I think this PR doesn't need to be perfect

agree.

Thegaram (approved):

PR #13, opened 2024-06-25
fix: add more validation checks

colinlyguo:
### PR title

- [ ] fix: A bug fix

### Breaking change label


georgehao (approved):

encoding/codecv0/codecv0.go
```diff
@@ -141,6 +141,14 @@ func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DACh
 
 // Encode serializes the DAChunk into a slice of bytes.
 func (c *DAChunk) Encode() ([]byte, error) {
+	if len(c.Blocks) == 0 {
```
zimpha, line 144: If checked in `NewDAChunk`, why do another check here?

colinlyguo: I think it's not worse to also add a check here. since DAChunk may be initialized by other means in unit tests.

zimpha (approved):

PR #16, opened 2024-07-06
fix: "small tip" block issue

noel2004:
zstd would, at least at our compress level, divide the input bytes into blocks, and compress each input block respectively.

The size of each input block is not exceed the MAX_BLOCK_SIZE (128kB), it can also be set with an experiment argument `max_block_size`.

Considering zstd has consume several blocks with `max_block_size` and there are only very few (like < 100) bytes left. The last block may be encoded as an `RAW` or even `RLE` block. Our decompress circuit would not support these types of blocks and the corresponding blob data would be fail to be finalized.

Such an incompatible blob data has caused the [incident in Jul. 5th](https://status.scroll.io/incidents/44k6s4qg6kcs)

Before this PR, we only set the "target block size" option, which has no effect on the size of input block (it only constraint the size of the output, i.e. compressed block). In this PR, our zstd encoder has applied the `max_block_size` correctly.

In this PR, zstd encoder would "average" the input block size used by zstd when the total input bytes has exceed 128kB, i.e. they can not be hold within a single block. With this "average" trick each input block would has enough bytes to be encoded by the normal "compressed" type.

We also induce a new entry from libzstd as `compress_scroll_batch_bytes_ex`, which enable the caller to control input block size. So we could tune the block size in golang side later.

To make more reliability, we induced a `CheckCompressedDataCompatibility` method in encoding module. ccc can apply this method when the uncompressed batch data has exceed 128kB **(important: never apply it when data is less than 128kB)**. If this fast checking fail, ccc can act like when the incremement data has broken any limits (drop the current data and seal the batch).


encoding/codecv2/codecv2_test.go
```diff
@@ -883,6 +883,33 @@ func TestCodecV2ChunkAndBatchCalldataSizeEstimation(t *testing.T) {
...
+
+	// magics := []byte{0x28, 0xb5, 0x2f, 0xfd}
+	// compressedBytes := append(magics, compressed...)
+
+	// err = os.WriteFile("../testdata/bug_out.zstd", compressedBytes, 0644)
+	// assert.NoError(t, err)
+
+	// decompressed, err := decompressScrollBatchBytes(compressedBytes)
+	// assert.NoError(t, err)
+	// assert.Equal(t, blobBytes, decompressed)
```
colinlyguo, line 910: Why commenting out these assertions?

noel2004: some code from the `feat/add-decoding-methods` branch but we still have no decompress routine in `main`

encoding/codecv2/libscroll_zstd_linux_amd64.a
colinlyguo, line 1: do we need to update other .a files as well?

noel2004: yes. Please

colinlyguo: Yes. this will fail in mac:
```
    codecv2_test.go:900: 
                Error Trace:    /Users/colin/Desktop/桌面/scroll_git/da-codec/encoding/codecv2/codecv2_test.go:900
                Error:          Received unexpected error:
                                unexpected blk type {0}, size {64}, last {true}
                Test:           TestBugRoundTripData
--- FAIL: TestBugRoundTripData (0.04s)
FAIL
FAIL    github.com/scroll-tech/da-codec/encoding/codecv2        6.136s
```

colinlyguo (approved):

encoding/da.go
```diff
@@ -269,3 +269,54 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
 	}
 	return txsData
 }
+
+// Fast testing if the compressed data is compatible with our circuit
+// (require specified frame header and each block is compressed)
+func TestCompressedDataCompatibility(data []byte) error {
```
Thegaram, line 275: I don't fully understand how we are supposed to use `TestCompressedDataCompatibility`.

You mentioned ccc, but we don't run ccc inside `rollup-relayer` (implementation detail).

But it sounds like you suggest calling this function on `compressedBlobBytes` (that we produce [here](https://github.com/scroll-tech/da-codec/blob/5b6cded48ab789c6c7210f113bbea900c32734e2/encoding/codecv2/codecv2.go#L181)), is that correct?

Thegaram (commented):
The PR makes sense. I'll do another review pass tomorrow.

encoding/da.go
noel2004: Yes. Or more specificied, in `EstimateBatchL1CommitBatchSizeAndBlobSize`?

If the test had failed, it can be handled just like the blobBytes has exceed the limit

HAOYUatHZ, line 275: prefer renaming it to `CheckCompressedDataCompatibility`

because in golang, functions named `TestXXXX` usually indicate test code.

libzstd/src/lib.rs
```diff
@@ -29,7 +29,21 @@ pub unsafe extern "C" fn compress_scroll_batch_bytes(
...
+    // when the src size larger than one input block (so the output has too
+    // be distributed in multiple blocks), we average each input block to
+    // avoid the small size in last input, which may lead to Raw or RLE blocks
+    let mut encoder = if src_size > N_BLOCK_SIZE_TARGET as u64{
+        let exp_blocks = (src_size - 1) / N_BLOCK_SIZE_TARGET as u64 + 1;
+        /* 
+            slightly increase for blocks except the last one,
+            so ensure no "tip" would be left
+        */
+        let reset_blk_size = src_size / exp_blocks + 8;
```
HAOYUatHZ, line 41: Is this safe?

so basically what we had is 

(`src_size` - 1) / (`exp_blocks` - 1 ) = `N_BLOCK_SIZE_TARGET`


now we want to pick a `reset_blk_size` so that 


(`src_size` - 1) / (`reset_exp_blocks` - 1 ) != reset_blk_size


but if `reset_blk_size = src_size / exp_blocks + 8`

as you can see, 
this equation is still possible

QianJianTec1720365273853

encoding/da.go
noel2004: Yes i have forgot about that ...

libzstd/src/lib.rs
noel2004: Could you explain where the `reset_exp_blocks` comes from?

noel2004:
> random question:
> 
> do we have a way to verify that the artifacts are generated by the corresponding codes? so no one injects some malicious codes?

You mean the `.a` stastic libs? In fact they can be generate by building from the rust code with a simple `cargo build`. I think we can warn everyone to generate them by theirself if possible instead of linking these artifacts

libzstd/src/lib.rs
HAOYUatHZ: > Could you explain where the `reset_exp_blocks` comes from?

similar to `exp_blocks` = `src_size` / `N_BLOCK_SIZE_TARGET`,
 `reset_exp_blocks` is `src_size` / `reset_blk_size`

HAOYUatHZ:
> > random question:
> > do we have a way to verify that the artifacts are generated by the corresponding codes? so no one injects some malicious codes?
> 
> You mean the `.a` stastic libs? In fact they can be generate by building from the rust code with a simple `cargo build`. I think we can warn everyone to generate them by theirself if possible instead of linking these artifacts

yeah but in https://github.com/scroll-tech/go-ethereum & https://github.com/scroll-tech/scroll we directly link & use these binaries,
which means if a dev build from malicious codes, we will deploy the malicious codes.

I think we can use a github action:
```
(in a same PR)
if a new commit pushed to github:
    build the binaries
    check MD5 change
        if MD5 changed:
            commit the new files and push to github
```

This can address my security concern but it's not related to this PR, we can do it later.

HAOYUatHZ:
I re-read everything again, now I have another question:

`let reset_blk_size = src_size / exp_blocks + 8;` so basically we want `reset_blk_size` at least `> 8`?

noel2004:
> I re-read everything again, now I have another question regarding `let reset_blk_size = src_size / exp_blocks + 8;`:
> 
> so basically we want `reset_blk_size` at least `> 8`?

Yes. The `reset_blk_size` is expected to be in range of (64kB, 128kB). And **every** block actually being compressed would has a size closed to `reset_blk_size`.

Of course if we have involved thousands of block, the "+8" trick would cause the final block become small again but it is impossible at our current scheme.

encoding/da.go
```diff
@@ -269,3 +269,54 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
 	}
 	return txsData
 }
+
+// Fast testing if the compressed data is compatible with our circuit
+// (require specified frame header and each block is compressed)
+func CheckCompressedDataCompatibility(data []byte) error {
```
colinlyguo, line 275: Do we also need to add this check after trying to compress every batch for early return?

encoding/da.go
```diff
@@ -269,3 +269,54 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
...
+		panic("impossible")
+	}
+
+	isLast := false
+	// scan each block until done
+	for len(data) > 3 && !isLast {
+		isLast = (data[0] & 1) == 1
+		blkType := (data[0] >> 1) & 3
+		blkSize := (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
+		//println("blk", blkType, blkSize, isLast)
```
colinlyguo, line 306:
```suggestion
```

encoding/da.go
```diff
@@ -269,3 +269,54 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
...
+		if blkType != 2 {
+			return fmt.Errorf("unexpected blk type {%d}, size {%d}, last {%t}", blkType, blkSize, isLast)
+		}
+		if len(data) < 3+int(blkSize) {
+			return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
+		}
+		data = data[3+blkSize:]
+	}
+
+	// should we wrong is isLast is still false?
```
colinlyguo, line 316:
```suggestion
	// Should we return invalid if isLast is still false?
```

encoding/da.go
noel2004: As mentioned above:

>can apply this method when the uncompressed batch data has exceed 128kB (important: never apply it when data is less than 128kB).

When the size of all uncompressed batches exceed 128kB, apply this check would bring an additional ensurance

libzstd/src/lib.rs
```diff
@@ -29,7 +29,21 @@ pub unsafe extern "C" fn compress_scroll_batch_bytes(
...
+        assert!(reset_blk_size <= N_BLOCK_SIZE_TARGET as u64);
```
colinlyguo, line 42: if `src_size == 2 * N_BLOCK_SIZE_TARGET - 2`, looks like it will break the assertion in this line.

libzstd/encoder/src/lib.rs
```diff
@@ -9,14 +9,14 @@ pub use zstd;
 #[allow(dead_code)]
 pub const CL_WINDOW_LIMIT: usize = 17;
 
-/// zstd block size target.
-pub const N_BLOCK_SIZE_TARGET: u32 = 124 * 1024;
+/// default targetted block size is max block size (128kB).
```
colinlyguo, line 12:
```suggestion
/// default targeted block size is max block size (128kB).
```

encoding/da.go
colinlyguo: so could i add this check in this pr?

colinlyguo: added in [01c9be9](https://github.com/scroll-tech/da-codec/pull/16/commits/01c9be9125aad91d0052fafdd4b76ad5cb8490cf), pls review, `codecv3` also uses this function.

colinlyguo:
unit tests run into this after adding the check (`CheckCompressedDataCompatibility`):
```
                Error:          Received unexpected error:
                                too small size (20b6550000180001000100300a0c01), what is it?
                Test:           TestCodecV3BatchEncode
```
I think `len(data) < 16` is possible if the blocks in the batch only contain L1 messages.

noel2004:
> unit tests run into this after adding the check (`CheckCompressedDataCompatibility`):
> 
> ```
>                 Error:          Received unexpected error:
>                                 too small size (20b6550000180001000100300a0c01), what is it?
>                 Test:           TestCodecV3BatchEncode
> ```
> 
> I think `len(data) < 16` is possible if the blocks in the batch only contain L1 messages. the compressed data length would be < 16.

That would be possible. It is the reason that i have emphased mulitple times that this checking **should not be applied when data is less than 128kB**

colinlyguo:
> > unit tests run into this after adding the check (`CheckCompressedDataCompatibility`):
> > ```
> >                 Error:          Received unexpected error:
> >                                 too small size (20b6550000180001000100300a0c01), what is it?
> >                 Test:           TestCodecV3BatchEncode
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think `len(data) < 16` is possible if the blocks in the batch only contain L1 messages. the compressed data length would be < 16.
> 
> That would be possible. It is the reason that i have emphased mulitple times that this checking **should not be applied when data is less than 128kB**

sorry, will add this check.

colinlyguo:
> 8cb5fe0

added in [8cb5fe0](https://github.com/scroll-tech/da-codec/pull/16/commits/8cb5fe07d64041d9dc19ac2a3b3337ab18b06ea0). the PR's description has not defined the case when uncompressed data length = 128KiB. so I use > 128KiB as the condition, feel free to tweak it.

colinlyguo:
> > unit tests run into this after adding the check (`CheckCompressedDataCompatibility`):
> > ```
> >                 Error:          Received unexpected error:
> >                                 too small size (20b6550000180001000100300a0c01), what is it?
> >                 Test:           TestCodecV3BatchEncode
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think `len(data) < 16` is possible if the blocks in the batch only contain L1 messages. the compressed data length would be < 16.
> 
> That would be possible. It is the reason that i have emphased mulitple times that this checking **should not be applied when data is less than 128kB**

added in [8cb5fe0](https://github.com/scroll-tech/da-codec/pull/16/commits/8cb5fe07d64041d9dc19ac2a3b3337ab18b06ea0). the PR's description has not defined the case when uncompressed data length = 128KiB. so I use > 128KiB as the condition, feel free to tweak it.

colinlyguo (dismissed):

libzstd/src/lib.rs
noel2004: It is the case. I think it should be ok to drop this "add small tip" trick.

noel2004: It is the case. I think it should be ok to drop this "add small tip" trick.

Thegaram (dismissed):

encoding/codecv2/codecv2.go
```diff
@@ -183,6 +184,14 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 		return nil, common.Hash{}, nil, err
 	}
 
+	// Only apply this check when the uncompressed batch data has exceeded 128KiB.
+	if len(blobBytes) > 131072 {
+		// check compressed data compatibility
+		if err = encoding.CheckCompressedDataCompatibility(compressedBlobBytes); err != nil {
+			return nil, common.Hash{}, nil, err
```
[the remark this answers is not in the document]
colinlyguo: I think we can embed it here since it's encoding's check instead of batch-proposer.
btw, is there anything suspicious about batch-proposer? maybe I have ignored it.

encoding/codecv2/codecv2_test.go
```diff
@@ -883,6 +883,33 @@ func TestCodecV2ChunkAndBatchCalldataSizeEstimation(t *testing.T) {
 	assert.Equal(t, uint64(180), batch5CalldataSize)
 }
 
+func TestBugRoundTripData(t *testing.T) {
```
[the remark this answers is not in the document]
noel2004: Simply take the name from here:
https://github.com/scroll-tech/zkevm-circuits/blob/99933d90d5cda9a9b634ee35cad3b8207fdd9d62/aggregator/src/aggregation/decoder/witgen.rs#L1910

encoding/codecv2/codecv2.go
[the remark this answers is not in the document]
colinlyguo: discussed with @noel2004, will apply this fix: https://github.com/scroll-tech/da-codec/pull/16#discussion_r1667594324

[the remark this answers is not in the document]
colinlyguo: moved the check to `EstimateChunkL1CommitBatchSizeAndBlobSize` and `EstimateBatchL1CommitBatchSizeAndBlobSize` in this commit: [a554fe4](https://github.com/scroll-tech/da-codec/pull/16/commits/a554fe417a393500de4f7545c6a72bb10a3351ff), pls review. also cc @noel2004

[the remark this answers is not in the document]
Thegaram: OK, looks good to me. If you're confident that adding this check here ensures that all chunks and batches pass the `CheckCompressedDataCompatibility` check, then this should be good.

colinlyguo (approved):

encoding/codecv2/codecv2.go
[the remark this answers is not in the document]
noel2004: So again i emphased that we only need this check when uncompressed bytes exceed 128kB

encoding/da.go
```diff
@@ -269,3 +269,53 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
...
+	case 0:
+		data = data[2:]
+	case 1:
+		data = data[3:]
+	case 2:
+		data = data[5:]
+	case 3:
+		data = data[9:]
+	default:
+		panic("impossible")
```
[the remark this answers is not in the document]
georgehao: lib should not throw panic out. use error instead

georgehao (approved):

Thegaram (commented):
Close in favor of https://github.com/scroll-tech/da-codec/pull/17?

PR #17, opened 2024-07-09
fix: check compressed data compatibility

colinlyguo:

Temporary fix, backward compatible.

rollup-relayer will handle this special error and early stop adding more blocks into a chunk (and more chunks into a batch).

### PR title

- [x] fix: A bug fix

### Breaking change label

### DIFF
--- a/encoding/codecv2/codecv2.go
+++ b/encoding/codecv2/codecv2.go
@@ -124,7 +124,7 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 	metadataLength := 2 + MaxNumChunks*4
 
 	// the raw (un-compressed and un-padded) blob payload
-	batchBytes := make([]byte, metadataLength)
+	blobBytes := make([]byte, metadataLength)
 
 	// challenge digest preimage
 	// 1 hash for metadata, 1 hash for each chunk, 1 hash for blob versioned hash
@@ -134,12 +134,12 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 	var chunkDataHash common.Hash
 
 	// blob metadata: num_chunks
-	binary.BigEndian.PutUint16(batchBytes[0:], uint16(len(chunks)))
+	binary.BigEndian.PutUint16(blobBytes[0:], uint16(len(chunks)))
 
 	// encode blob metadata and L2 transactions,
 	// and simultaneously also build challenge preimage
 	for chunkID, chunk := range chunks {
-		currentChunkStartIndex := len(batchBytes)
+		currentChunkStartIndex := len(blobBytes)
 
 		for _, block := range chunk.Blocks {
 			for _, tx := range block.Transactions {
@@ -152,17 +152,17 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 				if err != nil {
 					return nil, common.Hash{}, nil, err
 				}
-				batchBytes = append(batchBytes, rlpTxData...)
+				blobBytes = append(blobBytes, rlpTxData...)
 			}
 		}
 
 		// blob metadata: chunki_size
-		if chunkSize := len(batchBytes) - currentChunkStartIndex; chunkSize != 0 {
-			binary.BigEndian.PutUint32(batchBytes[2+4*chunkID:], uint32(chunkSize))
+		if chunkSize := len(blobBytes) - currentChunkStartIndex; chunkSize != 0 {
+			binary.BigEndian.PutUint32(blobBytes[2+4*chunkID:], uint32(chunkSize))
 		}
 
 		// challenge: compute chunk data hash
-		chunkDataHash = crypto.Keccak256Hash(batchBytes[currentChunkStartIndex:])
+		chunkDataHash = crypto.Keccak256Hash(blobBytes[currentChunkStartIndex:])
 		copy(challengePreimage[32+chunkID*32:], chunkDataHash[:])
 	}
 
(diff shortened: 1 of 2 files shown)


noel2004 (approved):

encoding/codecv2/codecv2.go
```diff
@@ -305,6 +306,13 @@ func EstimateChunkL1CommitBatchSizeAndBlobSize(c *encoding.Chunk) (uint64, uint6
 	if err != nil {
 		return 0, 0, err
 	}
+	// Only apply this check when the uncompressed batch data has exceeded 128 KiB.
+	if len(batchBytes) > 131072 {
+		// Check compressed data compatibility.
+		if err = encoding.CheckCompressedDataCompatibility(blobBytes); err != nil {
+			return math.MaxUint32, math.MaxUint32, nil // Return math.MaxUint32 to indicate the batch is too large and avoid overflows.
```
Thegaram, line 313:
```suggestion
			return math.MaxUint32, math.MaxUint32, nil // Return math.MaxUint32 to indicate the chunk is too large and avoid overflows.
```

encoding/codecv2/codecv2.go
```diff
@@ -318,6 +326,13 @@ func EstimateBatchL1CommitBatchSizeAndBlobSize(b *encoding.Batch) (uint64, uint6
 	if err != nil {
 		return 0, 0, err
 	}
+	// Only apply this check when the uncompressed batch data has exceeded 128 KiB.
+	if len(batchBytes) > 131072 {
+		// Check compressed data compatibility.
+		if err = encoding.CheckCompressedDataCompatibility(blobBytes); err != nil {
```
Thegaram, line 332: Let's add logs + metrics to these 2 error cases so that we can see how often they happen.

Thegaram (commented):
Looks good, let's also wait for @roynalnaruto's review before merging.

encoding/codecv2/codecv2.go
Thegaram: We can also add these on the caller side (in `chunk-proposer` and `batch-proposer`), but in that case it's better to return an error here. Your call @colinlyguo

encoding/da.go
```diff
@@ -269,3 +269,53 @@ func TxsToTxsData(txs types.Transactions) []*types.TransactionData {
...
+	case 0:
+		data = data[2:]
+	case 1:
+		data = data[3:]
+	case 2:
+		data = data[5:]
+	case 3:
+		data = data[9:]
+	default:
+		panic("impossible")
```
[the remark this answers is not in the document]
noel2004: fheader is a `byte` (i.e uint8) and there is impossible for a unit8 >> 6 obtain value beyond 0-3.

So it should be safe to panic here (like the unreachable block in a function)

encoding/codecv2/codecv2.go
colinlyguo: add a new error type (`CompressedDataCompatibilityError`) for `chunk-proposer` and `batch-proposer` to catch, will submit metrics in `chunk-proposer` and `batch-proposer`.

encoding/codecv2/codecv2.go
```diff
@@ -305,6 +306,14 @@ func EstimateChunkL1CommitBatchSizeAndBlobSize(c *encoding.Chunk) (uint64, uint6
...
+			log.Error("EstimateChunkL1CommitBatchSizeAndBlobSize: compressed data compatibility check failed", "err", err, "batchBytes", hex.EncodeToString(batchBytes), "blobBytes", hex.EncodeToString(blobBytes))
```
Thegaram, line 313:
```suggestion
			log.Warn("EstimateChunkL1CommitBatchSizeAndBlobSize: compressed data compatibility check failed", "err", err, "batchBytes", hex.EncodeToString(batchBytes), "blobBytes", hex.EncodeToString(blobBytes))
```

Thegaram (approved):

encoding/codecv2/codecv2.go
colinlyguo: changed in [3c6edd8](https://github.com/scroll-tech/da-codec/pull/17/commits/3c6edd82ce0f68a88a60f7baccb2aee6b5a6cd00).

Thegaram (approved):

noel2004 (approved):

encoding/codecv2/codecv2.go
```diff
@@ -183,6 +184,15 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
...
 	}
 
+	// Only apply this check when the uncompressed batch data has exceeded 128 KiB.
+	if len(blobBytes) > 131072 {
+		// Check compressed data compatibility.
+		if err = encoding.CheckCompressedDataCompatibility(blobBytes); err != nil {
+			log.Error("ConstructBlobPayload: compressed data compatibility check failed", "err", err, "uncompressedBlobBytes", hex.EncodeToString(blobBytes), "compressedBlobBytes", hex.EncodeToString(compressedBlobBytes))
+			return nil, common.Hash{}, nil, &encoding.CompressedDataCompatibilityError{Err: err}
+		}
+	}
```
roynalnaruto, line 194: @colinlyguo when does this scenario occur? `blobBytes < 4096 * 31` will always hold, right? So isn't this check redundant?

We should instead have this check for `batchBytes`, which is already done [here](https://github.com/scroll-tech/da-codec/pull/17/files#diff-626a61d73695a9154e13d35f43a9633078eacc8ce3e1cb8a290e8c0723a0a0dcR319-R325)

encoding/codecv2/codecv2.go
```diff
@@ -183,6 +184,15 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 		return nil, common.Hash{}, nil, err
 	}
 
+	// Only apply this check when the uncompressed batch data has exceeded 128 KiB.
+	if len(blobBytes) > 131072 {
```
roynalnaruto, line 188: This should be `len(batchBytes)` ?

roynalnaruto (commented):
May be an issue here?

encoding/codecv2/codecv2.go
colinlyguo: nice catch, this is fixed in https://github.com/scroll-tech/da-codec/pull/18. when using the new patch in rollup-relayer, some unit tests failed due to the bug you mentioned.

encoding/codecv2/codecv2.go
Thegaram: `blobBytes` here is uncompressed (as opposed to `compressedBlobBytes` a few lines above).

encoding/codecv2/codecv2.go
colinlyguo: blobBytes here means batchBytes, the naming is outdated.

PR #19, opened 2024-07-11
feat: add functions to check compressed data validity

colinlyguo:
### PR title

- [x] feat: A new feature

### Breaking change label


noel2004 (approved):
With this PR, da-codec only export `CheckBatchCompressedDataCompatibility` and do nothing about compatibility checking inside itself, correct?

colinlyguo:
> With this PR, da-codec only export `CheckBatchCompressedDataCompatibility` and do nothing about compatibility checking inside itself, correct?

It will do the check again when constructing the batches.

georgehao (approved):

Thegaram (approved):

PR #23, opened 2024-08-14
feat: support conditional compress

colinlyguo:

Added a flag `conditionalEncode` to enable rollup-relayer and l2geth-verifier to determine when to enable this.

### PR title

- [x] feat: A new feature

### Breaking change label


encoding/codecv2/codecv2.go
```diff
@@ -193,6 +193,20 @@ func ConstructBlobPayload(chunks []*encoding.Chunk, useMockTxData bool) (*kzg484
 		}
 	}
 
+	if conditionalEncode {
+		encoded := len(blobBytes) < len(batchBytes)
+		if encoded {
```
Thegaram, line 198: We should only compress and check compatibility if the caller requested `encoded = true`.

encoding/codecv2/codecv2.go
```diff
@@ -315,11 +329,15 @@ func EstimateChunkL1CommitBatchSizeAndBlobSize(c *encoding.Chunk) (uint64, uint6
 	if err != nil {
 		return 0, 0, err
 	}
-	return uint64(len(batchBytes)), CalculatePaddedBlobSize(uint64(len(blobBytes))), nil
+	blobBytesLen := uint64(len(blobBytes))
+	if conditionalEncode {
+		blobBytesLen += 1
+	}
```
Thegaram, line 335: I don't understand this, shouldn't it be something like this?

```go
if conditionalEncode {
    blobBytes, err := compressScrollBatchBytes(batchBytes)
    blobBytesLen = 1 + len(blobBytes)
} else {
    blobBytesLen = 1 + len(batchBytes)
}
```

encoding/codecv2/codecv2.go
colinlyguo: Yes. this is an outdated implementation. will refactor to let rollup-relayer decide if it needs to be encoded.

encoding/codecv2/codecv2.go
colinlyguo: yeah. this should be fixed to align with compression behavior. previously I thought the only change was: that in blob bytes, there is only one extra flag to be added (but still missing some details), will refactor it to the new version.

encoding/codecv4/codecv4.go
```diff
@@ -0,0 +1,534 @@
...
+	return codecv3.NewDABlock(block, totalL1MessagePoppedBefore)
+}
+
+// NewDAChunk creates a new DAChunk from the given encoding.Chunk and the total number of L1 messages popped before.
+func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DAChunk, error) {
+	return codecv3.NewDAChunk(chunk, totalL1MessagePoppedBefore)
+}
+
+// NewDABatch creates a DABatch from the provided encoding.Batch.
+func NewDABatch(batch *encoding.Batch, enableEncode bool) (*DABatch, error) {
```
zimpha, line 68: looks like `enableEncode` this variable actual means whether to enable zstd compression, right? Better to change the name to `enableCompress`.

colinlyguo: makes sense. changed in [c6af3bb](https://github.com/scroll-tech/da-codec/pull/23/commits/c6af3bbe7068da2b356509f2cb9eaf6c8d514bdf).

encoding/codecv4/codecv4_test.go
```diff
@@ -0,0 +1,837 @@
...
+func TestCodecV4BatchEncode(t *testing.T) {
+	// empty batch
+	batch := &DABatch{Version: uint8(encoding.CodecV4)}
+	encoded := hex.EncodeToString(batch.Encode())
+	assert.Equal(t, "04000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000", encoded)
+
+	trace2 := readBlockFromJSON(t, "../testdata/blockTrace_02.json")
+	chunk2 := &encoding.Chunk{Blocks: []*encoding.Block{trace2}}
+	originalBatch := &encoding.Batch{Chunks: []*encoding.Chunk{chunk2}}
+	batch, err := NewDABatch(originalBatch, true /* enble encode */)
```
Thegaram, line 219:
```suggestion
	batch, err := NewDABatch(originalBatch, true /* enable encode */)
```

encoding/codecv4/codecv4.go
```diff
@@ -0,0 +1,446 @@
...
+	return codecv3.NewDABlock(block, totalL1MessagePoppedBefore)
+}
+
+// NewDAChunk creates a new DAChunk from the given encoding.Chunk and the total number of L1 messages popped before.
+func NewDAChunk(chunk *encoding.Chunk, totalL1MessagePoppedBefore uint64) (*DAChunk, error) {
+	return codecv3.NewDAChunk(chunk, totalL1MessagePoppedBefore)
+}
+
+// NewDABatch creates a DABatch from the provided encoding.Batch.
+func NewDABatch(batch *encoding.Batch, enableCompress bool) (*DABatch, error) {
```
Thegaram, line 63: How does the caller set `enableCompress`? Try with `true` first and if it errors then try again with `false`?

Thegaram (dismissed):

encoding/codecv4/codecv4.go
colinlyguo: Yes.
in l2geth-verifier: try with `true` first and if it errors then try again with `false`.
in rollup-relayer: will ask `CheckChunkCompressedDataCompatibility` and `CheckBatchCompressedDataCompatibility`, with some strategies to determine whether or not to use `enableCompress`.

encoding/codecv4/codecv4_test.go
Thegaram: Please update all occurrences

colinlyguo: fixed in [7fd9672](https://github.com/scroll-tech/da-codec/pull/23/commits/7fd9672293445b11d96e7f5528cfae7b52135e70).

zimpha (approved):

Thegaram (approved):

PR #33, opened 2024-12-27
feat(CodecV7): add CodecV7 to support upgrade 5.2 Euclid phase2

jonastheis:
This PR adds support for `CodecV7` for the `EuclidV2` hardfork.

### PR title

- [ ] build: Changes that affect the build system or external dependencies (example scopes: yarn, eslint, typescript)
- [ ] ci: Changes to our CI configuration files and scripts (example scopes: vercel, github, cypress)
- [ ] docs: Documentation-only changes
- [x] feat: A new feature
- [ ] fix: A bug fix
- [ ] perf: A code change that improves performance
- [ ] refactor: A code change that doesn't fix a bug, or add a feature, or improves performance
- [ ] style: Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc)
- [ ] test: Adding missing tests or correcting existing tests

### Breaking change label

- [ ] Yes

## Summary by CodeRabbit

- **New Features**
  • Introduced a new data processing version that improves how blobs and batches are created and decoded.
  • Enhanced support for L1 message handling with optimized processing and stronger error detection.
  • Added new methods for encoding and decoding functionalities in the latest codec version.
  • Implemented a new interface for blob payload management.
  • Added additional methods for estimating calldata sizes and gas costs.

- **Tests**
  • Expanded test coverage to validate encoding/decoding, rolling hash computations, and configuration changes.
  • Introduced new test cases for codec versions and configurations.

- **Chores**
  • Updated key dependencies to further boost performance and stability.


encoding/codecv6.go
```diff
@@ -0,0 +1,77 @@
...
+	return nil, nil
+}
+
+func (d *DACodecV6) DecodeTxsFromBlob(blob *kzg4844.Blob, chunks []*DAChunkRawTx) error {
+	return nil
+}
+
+// TODO: add DecodeBlob to interface to decode the blob and transactions or reuse DecodeTxsFromBlob but only have a single "chunk" for all transactions in the batch?
+
+// TODO: which of the Estimate* functions are needed?
```
colinlyguo, line 62: It's used to give proof witnesses (from coordinator to provers in distributing tasks)

encoding/codecv6_types.go
```diff
@@ -0,0 +1,135 @@
...
+	"encoding/json"
+	"errors"
+	"fmt"
+
+	"github.com/scroll-tech/go-ethereum/common"
+	"github.com/scroll-tech/go-ethereum/crypto"
+	"github.com/scroll-tech/go-ethereum/crypto/kzg4844"
+)
+
+// daBatchV3 contains metadata about a batch of DAChunks.
```
colinlyguo, line 15:
```suggestion
// daBatchV6 contains metadata about a batch of DAChunks.
```

encoding/codecv6_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+	version := CodecVersion(data[daBatchOffsetVersion])
+	batchIndex := binary.BigEndian.Uint64(data[daBatchOffsetBatchIndex:daBatchV6OffsetBlobVersionedHash])
+	blobVersionedHash := common.BytesToHash(data[daBatchV6OffsetBlobVersionedHash:daBatchV6OffsetParentBatchHash])
+	parentBatchHash := common.BytesToHash(data[daBatchV6OffsetParentBatchHash:daBatchV6EncodedLength])
+
+	return newDABatchV6(version, batchIndex, parentBatchHash, blobVersionedHash, nil, nil, nil)
+}
+
+// Encode serializes the DABatchV3 into bytes.
```
colinlyguo, line 55:
```suggestion
// Encode serializes the DABatchV6 into bytes.
```

encoding/codecv6.go
colinlyguo: previously it's used here: https://github.com/scroll-tech/scroll/blob/v4.4.83/common/types/message/message.go#L56.
now it is not used and may be removed.

encoding/codecv6.go
```diff
@@ -0,0 +1,77 @@
...
+}
+
+// Version returns the codec version.
+func (d *DACodecV6) Version() CodecVersion {
+	return CodecV6
+}
+
+// MaxNumChunksPerBatch returns the maximum number of chunks per batch.
+func (d *DACodecV6) MaxNumChunksPerBatch() int {
+	return 1
```
Thegaram, line 21: Why 1? Just a placeholder?

encoding/codecv6.go
Thegaram, line 1: Probably need to change to v7 to accommodate https://github.com/scroll-tech/da-codec/pull/34

The plan is that v5 will mean "special MPT migration transition batch" (one-off), and v6 will mean "mpt batch". Other than this v4, v5, v6 will be identical.

encoding/codecv6_types.go
```diff
@@ -0,0 +1,135 @@
...
+	return batchBytes
+}
+
+// Hash computes the hash of the serialized DABatch.
+func (b *daBatchV6) Hash() common.Hash {
+	return crypto.Keccak256Hash(b.Encode())
+}
+
+// BlobDataProofForPointEvaluation computes the abi-encoded blob verification data.
+func (b *daBatchV6) BlobDataProofForPointEvaluation() ([]byte, error) {
```
Thegaram, line 71: Can be removed

encoding/codecv6_types.go
```diff
@@ -0,0 +1,135 @@
...
+
+// newDABatchV6 is a constructor for daBatchV6 that calls blobDataProofForPICircuit internally.
+func newDABatchV6(version CodecVersion, batchIndex uint64, parentBatchHash, blobVersionedHash common.Hash, blob *kzg4844.Blob, z *kzg4844.Point, blobBytes []byte) (*daBatchV6, error) {
+	daBatch := &daBatchV6{
+		version:           version,
+		batchIndex:        batchIndex,
+		parentBatchHash:   parentBatchHash,
+		blobVersionedHash: blobVersionedHash,
+		blob:              blob,
+		z:                 z,
```
Thegaram, line 35: `z` can be removed

encoding/codecv6_types.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// Version returns the version of the DABatch.
+func (b *daBatchV6) Version() CodecVersion {
+	return b.version
+}
+
+// SkippedL1MessageBitmap returns the skipped L1 message bitmap of the DABatch.
+// For daBatchV6, there is no skipped L1 message bitmap.
+func (b *daBatchV6) SkippedL1MessageBitmap() []byte {
```
Thegaram, line 127: Are these two just kept for interface consistency?

encoding/codecv6.go
jonastheis: Since `Chunks` are not exposed to DA, there should be no details about chunks here

encoding/codecv6_types.go
jonastheis: kept to fulfill interface

encoding/codecv6_types.go
jonastheis: yes but I'm not sure how we should deal with this right now. easiest is just to leave them and return empty values (or panic).

encoding/codecv6.go
Thegaram: In that case, `rollup-relayer` should not use `MaxNumChunksPerBatch`, right? Otherwise we'll keep producing batches with a single chunk.

encoding/codecv7_types.go
```diff
@@ -0,0 +1,353 @@
...
+func (b *daBatchV7) Blob() *kzg4844.Blob {
+	return b.blob
+}
+
+// BlobBytes returns the blob bytes of the batch.
+func (b *daBatchV7) BlobBytes() []byte {
+	return b.blobBytes
+}
+
+// MarshalJSON implements the custom JSON serialization for daBatchV3.
```
colinlyguo, line 118:
```suggestion
// MarshalJSON implements the custom JSON serialization for daBatchV7.
```

colinlyguo (approved):

encoding/codecv6.go
jonastheis: yeah might be better to return `maxInt` here instead. then we can configure a lower target in the relayer

encoding/codecv7.go
```diff
@@ -0,0 +1,319 @@
...
+	panic("implement me")
+}
+
+func (d *DACodecV7) EstimateBatchL1CommitCalldataSize(batch *Batch) (uint64, error) {
+	//TODO implement me after contracts are implemented
+	panic("implement me")
+}
+
+// JSONFromBytes converts the bytes to a DABatch and then marshals it to JSON.
+func (d *DACodecV7) JSONFromBytes(data []byte) ([]byte, error) {
```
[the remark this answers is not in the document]
jonastheis: It is part of the `Codec` interface. If you think we can remove it we can remove it from all versions.

encoding/codecv7_types.go
```diff
@@ -0,0 +1,353 @@
...
+
+	return res, nil
+}
+
+func decodeSize3Bytes(data []byte) uint32 {
+	return uint32(data[0]) | uint32(data[1])<<8 | uint32(data[2])<<16
+}
+
+func encodeSize3Bytes(data uint32) []byte {
+	return []byte{byte(data), byte(data >> 8), byte(data >> 16)}
```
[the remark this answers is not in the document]
jonastheis: yeah you're right it basically is little endian. however this is the custom defined byte size for the blob envelope. out of simplicity I'd even suggest just using a `uint32` here instead of the 3 custom bytes

encoding/codecv6.go
jonastheis: done in 1dde89ae3ab03889645a294ecb35afc549ceef1f

encoding/codecv7_types.go
```diff
@@ -0,0 +1,353 @@
...
+		return nil, fmt.Errorf("invalid data length for blobPayloadV7, expected at least %d bytes but got %d", blobPayloadV7EncodedLength, len(data))
+	}
+
+	initialL1MessageIndex := binary.BigEndian.Uint64(data[blobPayloadV7OffsetInitialL1MessageIndex:blobPayloadV7OffsetInitialL1MessageQueue])
+	initialL1MessageQueueHash := common.BytesToHash(data[blobPayloadV7OffsetInitialL1MessageQueue:blobPayloadV7OffsetLastL1MessageQueue])
+	lastL1MessageQueueHash := common.BytesToHash(data[blobPayloadV7OffsetLastL1MessageQueue:blobPayloadV7OffsetInitialL2BlockNumber])
+
+	initialL2BlockNumber := binary.BigEndian.Uint64(data[blobPayloadV7OffsetInitialL2BlockNumber:blobPayloadV7OffsetNumBlocks])
+	numBlocks := int(binary.BigEndian.Uint16(data[blobPayloadV7OffsetNumBlocks:blobPayloadV7OffsetBlocks]))
+
```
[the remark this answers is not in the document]
jonastheis: done in 1dde89ae3ab03889645a294ecb35afc549ceef1f

encoding/da.go
```diff
@@ -131,6 +137,19 @@ func (b *Block) NumL1Messages(totalL1MessagePoppedBefore uint64) uint64 {
 	return *lastQueueIndex - totalL1MessagePoppedBefore + 1
 }
 
+// NumL1MessagesNoSkipping returns the number of L1 messages in this block.
+// This method assumes that L1 messages can't be skipped.
+func (b *Block) NumL1MessagesNoSkipping() uint16 {
```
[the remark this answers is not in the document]
jonastheis: done in 1dde89ae3ab03889645a294ecb35afc549ceef1f

encoding/codecv7.go
```diff
@@ -0,0 +1,319 @@
...
+		uint16(numTransactions),      // numTransactions
+		numL1Messages,                // numL1Messages
+	)
+
+	return daBlock, nil
+}
+
+// NewDAChunk creates a new DAChunk from the given Chunk and the total number of L1 messages popped before.
+// Note: For DACodecV7, this function is not implemented since there is no notion of DAChunk in this version. Blobs
+// contain the entire batch data, and it is up to a prover to decide the chunk sizes.
```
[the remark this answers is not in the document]
jonastheis: adjusted in 1dde89ae3ab03889645a294ecb35afc549ceef1f

encoding/codecv7_types.go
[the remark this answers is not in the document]
Thegaram: Do you mean changing it to 4 bytes and using `binary.BigEndian.PutUint64`? Do you plan to change this?

encoding/codecv7.go
```diff
@@ -0,0 +1,319 @@
...
+	daBatch, err := newDABatchV7(CodecV7, batch.Index, blobVersionedHash, batch.ParentBatchHash, blob, blobBytes)
+	if err != nil {
+		return nil, fmt.Errorf("failed to construct DABatch: %w", err)
+	}
+
+	return daBatch, nil
+}
+
+func (d *DACodecV7) constructBlob(batch *Batch) (*kzg4844.Blob, common.Hash, []byte, error) {
+	enableCompression, err := d.CheckBatchCompressedDataCompatibility(batch)
```
Thegaram, line 93: Just noting that it seems we're calling `constructBlobPayload` twice? Once inside `CheckBatchCompressedDataCompatibility` and once a couple of lines below. (But this is not introduced in this PR.)

encoding/codecv7_types.go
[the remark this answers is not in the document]
Thegaram: Though it seems sufficient to reverse the order.

```suggestion
	return []byte{byte(data >> 16), byte(data >> 8), byte(data)}
```

encoding/codecv7.go
```diff
@@ -0,0 +1,319 @@
...
+		}
+		blobBytes[blobEnvelopeV7OffsetCompressedFlag] = 0x1
+		payloadBytes = compressedPayloadBytes
+	} else {
+		blobBytes[blobEnvelopeV7OffsetCompressedFlag] = 0x0
+	}
+
+	sizeSlice := encodeSize3Bytes(uint32(len(payloadBytes)))
+	copy(blobBytes[blobEnvelopeV7OffsetByteSize:blobEnvelopeV7OffsetCompressedFlag], sizeSlice)
+	blobBytes = append(blobBytes, payloadBytes...)
```
Thegaram, line 126: Small nitpick that it's simpler to follow if you prepare the data first then encode in one code block, in encoding order

```go
blobBytes[blobEnvelopeV7OffsetVersion] = uint8(CodecV7)
copy(blobBytes[blobEnvelopeV7OffsetByteSize:blobEnvelopeV7OffsetCompressedFlag], sizeSlice)
blobBytes[blobEnvelopeV7OffsetCompressedFlag] = isCompressed
blobBytes = append(blobBytes, payloadBytes...)
```

encoding/codecv7.go
```diff
@@ -0,0 +1,319 @@
...
+	rawBytes := bytesFromBlobCanonical(blob)
+
+	// read the blob envelope header
+	version := rawBytes[blobEnvelopeV7OffsetVersion]
+	if CodecVersion(version) != CodecV7 {
+		return nil, fmt.Errorf("codec version mismatch: expected %d but found %d", CodecV7, version)
+	}
+
+	// read the data size
+	blobEnvelopeSize := decodeSize3Bytes(rawBytes[blobEnvelopeV7OffsetByteSize:blobEnvelopeV7OffsetCompressedFlag])
```
Thegaram, line 193:
```suggestion
	blobPayloadSize := decodeSize3Bytes(rawBytes[blobEnvelopeV7OffsetByteSize:blobEnvelopeV7OffsetCompressedFlag])
```

encoding/da.go
```diff
@@ -132,6 +141,42 @@ func (b *Block) NumL1Messages(totalL1MessagePoppedBefore uint64) uint64 {
 	return *lastQueueIndex - totalL1MessagePoppedBefore + 1
 }
 
+// NumL1MessagesNoSkipping returns the number of L1 messages in this block.
```
Thegaram, line 144:
```suggestion
// NumL1MessagesNoSkipping returns the number of L1 messages and the highest queue index in this block.
```

encoding/da.go
```diff
@@ -714,3 +760,59 @@ func GetBatchEnableCompression(codecVersion CodecVersion, batch *Batch) (bool, e
...
+				return common.Hash{}, fmt.Errorf("failed to decode txData.Data: data=%v, err=%w", txData.Data, err)
+			}
+
+			l1Message := &types.L1MessageTx{
+				QueueIndex: txData.Nonce,
+				Gas:        txData.Gas,
+				To:         txData.To,
+				Value:      txData.Value.ToInt(),
+				Data:       data,
+				// Sender:     , TODO: is this needed?
```
Thegaram, line 783: Yes, needed, or the hashes won't match. This field is part of the [RLP encoding](https://github.com/scroll-tech/go-ethereum/blob/d2f19ff58eecb805ced826fc71da7521e9c2ea9b/core/types/l1_message_tx.go#L12).

```suggestion
				Sender:     txData.from,
```

encoding/da.go
```diff
@@ -714,3 +760,59 @@ func GetBatchEnableCompression(codecVersion CodecVersion, batch *Batch) (bool, e
...
+}
+
+func messageQueueV2ApplyL1Message(initialQueueHash common.Hash, message *types.L1MessageTx) common.Hash {
+	rollingHash := crypto.Keccak256Hash(initialQueueHash.Bytes(), types.NewTx(message).Hash().Bytes())
+
+	return messageQueueV2EncodeRollingHash(rollingHash)
+}
+
+func messageQueueV2EncodeRollingHash(rollingHash common.Hash) common.Hash {
+	// clear last 36 bits
```
Thegaram, line 809: This is 44 bits.

Maybe we should add some matching unit tests here and in the contracts PR

encoding/codecv7_types.go
```diff
@@ -0,0 +1,432 @@
...
+	blobPayloadV7OffsetInitialL1MessageQueue = 8
+	blobPayloadV7OffsetLastL1MessageQueue    = 40
+	blobPayloadV7OffsetInitialL2BlockNumber  = 72
+	blobPayloadV7OffsetNumBlocks             = 80
+	blobPayloadV7OffsetBlocks                = 82
+)
+
+// Below is the encoding for DABlockV7, total 52 bytes.
+//   * Field                   Bytes      Type         Index  Comments
+//   * blockNumber             8          uint64       0      The height of this block.
```
Thegaram, line 71:
```suggestion
```

encoding/codecv7_types.go
```diff
@@ -0,0 +1,432 @@
...
+
+func (b *blobPayloadV7) Encode() ([]byte, error) {
+	payloadBytes := make([]byte, blobPayloadV7MinEncodedLength)
+
+	binary.BigEndian.PutUint64(payloadBytes[blobPayloadV7OffsetInitialL1MessageIndex:blobPayloadV7OffsetInitialL1MessageQueue], b.initialL1MessageIndex)
+	copy(payloadBytes[blobPayloadV7OffsetInitialL1MessageQueue:blobPayloadV7OffsetLastL1MessageQueue], b.initialL1MessageQueueHash[:])
+	copy(payloadBytes[blobPayloadV7OffsetLastL1MessageQueue:blobPayloadV7OffsetInitialL2BlockNumber], b.lastL1MessageQueueHash[:])
+
+	var initialL2BlockNumber uint64
+	if len(b.blocks) > 0 {
```
Thegaram, line 230: Is an empty batch (0 blocks) valid? If yes then let's sure to add a test for it

Thegaram (commented):
LGTM, let's merge once the rolling hash is fixed.

encoding/codecv7_types.go
jonastheis: there's a check to not allow for empty batch creation via the codec which is the only way to publicly interact: https://github.com/scroll-tech/da-codec/blob/b71c047ffecc91720cd8110f9fac227c84ef3e29/encoding/codecv7.go#L75

However, I added this check more from a user's PoV that we not accidentally create empty batches. I think technically it should be okay to create empty batches. in that case the blob payload will only contain the metadata

encoding/codecv7_types.go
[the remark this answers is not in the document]
jonastheis: done in cbed8b2507be70f82e6cb4663668c56ed0c72edf

encoding/codecv7.go
jonastheis: yep it was a remnant from how it worked before. You're right, I also don't see a reason to construct the blob twice and compress it twice. Changed in b71c047ffecc91720cd8110f9fac227c84ef3e29

encoding/codecv7.go
jonastheis: Done in cbed8b2507be70f82e6cb4663668c56ed0c72edf

encoding/da.go
jonastheis: done in cbed8b2507be70f82e6cb4663668c56ed0c72edf

encoding/da.go
jonastheis: done in cbed8b2507be70f82e6cb4663668c56ed0c72edf

encoding/codecv7_types.go
jonastheis: For `codecV7.NewDABatch(batch)`: Empty Batch, creation error=no blocks -> is tested [here](https://github.com/scroll-tech/da-codec/blob/cbed8b2507be70f82e6cb4663668c56ed0c72edf/encoding/codecv7_test.go#L120) and [here](https://github.com/scroll-tech/da-codec/blob/cbed8b2507be70f82e6cb4663668c56ed0c72edf/encoding/codecv7_test.go#L212)

jonastheis:
> let's merge once the rolling hash is fixed

The `Estimate*` functions are not yet implemented. I'd like to wait for the contract implementation to add those.

encoding/codecv7_types.go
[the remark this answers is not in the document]
Thegaram: Thanks. Then you also need to update the `BlobEnvelopeV7` docs

encoding/codecv7_types.go
Thegaram: I agree, technically empty batch is OK but we can disallow it.

encoding/codecv7_types.go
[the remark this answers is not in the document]
jonastheis: good point. I updated the spec but forgot it here. done in 392b6ff736d03d0bd35103a9482c5d5f1a026cbb

encoding/codecv7.go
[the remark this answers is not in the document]
colinlyguo: I think so. seems this interface is not used anywhere now.

encoding/codecv7.go
```diff
@@ -0,0 +1,351 @@
...
+
+	return daBatch, nil
+}
+
+func (d *DACodecV7) NewDABatchFromParams(batchIndex uint64, blobVersionedHash, parentBatchHash common.Hash) (DABatch, error) {
+	return newDABatchV7(CodecV7, batchIndex, blobVersionedHash, parentBatchHash, nil, nil)
+}
+
+func (d *DACodecV7) DecodeDAChunksRawTx(_ [][]byte) ([]*DAChunkRawTx, error) {
+	return nil, nil
```
Thegaram, line 170: Consider returning an error

encoding/codecv7.go
```diff
@@ -0,0 +1,351 @@
...
+		return 0, 0, fmt.Errorf("failed to check batch compressed data compatibility: %w", err)
+	}
+
+	if enableCompression {
+		blobBytes = append(blobBytes, compressedPayloadBytes...)
+	} else {
+		blobBytes = append(blobBytes, payloadBytes...)
+	}
+
+	return blobEnvelopeV7OffsetPayload + uint64(len(payloadBytes)), calculatePaddedBlobSize(uint64(len(blobBytes))), nil
```
Thegaram, line 283: This seems correct, but now the concerpts of "L1 commit batch size" and "blob size" are quite confusing

encoding/codecv7.go
```diff
@@ -0,0 +1,351 @@
...
+func (d *DACodecV7) EstimateChunkL1CommitGas(chunk *Chunk) (uint64, error) {
+	return 0, nil
+}
+
+// EstimateBatchL1CommitGas calculates the total L1 commit gas for this batch approximately.
+func (d *DACodecV7) EstimateBatchL1CommitGas(batch *Batch) (uint64, error) {
+	// TODO: adjust this after contracts are implemented
+	var totalL1CommitGas uint64
+
+	// Add extra gas costs
```
Thegaram, line 328: We'd also have some hashing (`computeBatchHash`). But it seems unnecessary to use the exact formula here since it's unlikely that we go over the block gas limit.

Thegaram (approved):

encoding/codecv7.go
colinlyguo: more context here: https://gist.github.com/roynalnaruto/3b02b468a9795fe7e830fb5e99c32bed

jonastheis: Agree the concept is a bit confusing. But I left it to be compatible with how we use it for previous versions and within the relayer. After the hard fork has happened we could simplify the relayer again and with it the `Codec` interface.

encoding/codecv7.go
jonastheis: yep I think a rough estimate should be fine here

encoding/codecv7.go
jonastheis: done in 75229310413548d117d864e493146a78b5938095

encoding/codecv7.go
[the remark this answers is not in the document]
jonastheis: Let's leave it for now and remove it with a cleanup PR after the hard fork

encoding/codecv7_types.go
```diff
@@ -0,0 +1,478 @@
...
+
+// Hash computes the hash of the serialized DABatch.
+func (b *daBatchV7) Hash() common.Hash {
+	return crypto.Keccak256Hash(b.Encode())
+}
+
+// BlobDataProofForPointEvaluation computes the abi-encoded blob verification data.
+// Note: This method is not implemented for daBatchV7.
+func (b *daBatchV7) BlobDataProofForPointEvaluation() ([]byte, error) {
+	return nil, nil
```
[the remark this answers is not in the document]
jonastheis: Done in 32f5b49b9e1940475eb4283e34a1366726292eee. After talking with @roynalnaruto, the function `BlobDataProofForPointEvaluation` will still be needed as we pass the kzg proof as witness to the circuit. Therefore, we still need to compute it and make it accessible from the relayer. This is done in 0247443faabd059fbbf12cf7da08793cc0da96c4

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+		}
+		// sanity check: L1 message indices are contiguous across blocks boundaries
+		if numL1Messages > 0 {
+			if l1MessageIndex+uint64(numL1Messages) != highestQueueIndex+1 {
+				return nil, fmt.Errorf("failed to sanity check L1 messages count after block %d: l1MessageIndex + numL1Messages != highestQueueIndex+1: %d + %d != %d", block.Header.Number.Uint64(), l1MessageIndex, numL1Messages, highestQueueIndex+1)
+			}
+			l1MessageIndex += uint64(numL1Messages)
+		}
+
+		daBlock := newDABlockV7(block.Header.Number.Uint64(), block.Header.Time, block.Header.BaseFee, block.Header.GasLimit, uint16(len(block.Transactions)), numL1Messages)
```
Thegaram, line 105: Why not reuse `NewDABlock` here? Is it because you want to avoid calling `NumL1MessagesNoSkipping` twice?

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+		return nil, common.Hash{}, nil, fmt.Errorf("failed to create blob commitment: %w", err)
+	}
+	blobVersionedHash := kzg4844.CalcBlobHashV1(sha256.New(), &c)
+
+	return blob, blobVersionedHash, blobBytes, nil
+}
+
+func (d *DACodecV7) constructBlobPayload(batch *Batch) ([]byte, error) {
+	blobPayload := blobPayloadV7{
+		initialL1MessageIndex:  batch.InitialL1MessageIndex,
```
Thegaram, line 215: TBD [discussion](https://scrollco.slack.com/archives/C07FNRFENCE/p1739896174288889?thread_ts=1739879623.366149&cid=C07FNRFENCE) about `initialL1MessageIndex`

jonastheis: as discussed removed it in 2043787704b00632d9de2a0ea3e7b0f9b0de21b3

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+		blocks = append(blocks, daBlock)
+		txs = append(txs, block.Transactions)
+	}
+
+	daChunk := newDAChunkV7(
+		blocks, // blocks
+		txs,    // transactions
+	)
+
+	// sanity check: prevL1MessageQueueHash+apply(L1Messages) = postL1MessageQueueHash
```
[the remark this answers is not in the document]
jonastheis: done in de09af4bc445412cf6fb4933a0213143b54b8b23

encoding/da.go
```diff
@@ -210,6 +260,7 @@ func convertTxDataToRLPEncoding(txData *types.TransactionData) ([]byte, error) {
 		})
 
 	case types.L1MessageTxType: // L1MessageTxType is not supported
+		fallthrough
 	default:
```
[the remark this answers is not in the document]
jonastheis: done inde09af4bc445412cf6fb4933a0213143b54b8b23

encoding/codecv7.go
jonastheis: Yes I wanted to avoid that. But I think we don't actually need to call it again. Changed in de09af4bc445412cf6fb4933a0213143b54b8b23 to use `NewDABlock`

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+		return nil, errors.New("block number is not uint64")
+	}
+
+	numL1Messages, highestQueueIndex, err := block.NumL1MessagesNoSkipping()
+	if err != nil {
+		return nil, fmt.Errorf("failed to calculate number of L1 messages: %w", err)
+	}
+	if totalL1MessagePoppedBefore+uint64(numL1Messages) != highestQueueIndex {
+		return nil, fmt.Errorf("failed to sanity check L1 messages count: totalL1MessagePoppedBefore + numL1Messages != highestQueueIndex: %d + %d != %d", totalL1MessagePoppedBefore, numL1Messages, highestQueueIndex)
+	}
```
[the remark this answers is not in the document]
jonastheis: Yep you are correct, didn't update it here. Done in de09af4bc445412cf6fb4933a0213143b54b8b23

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+
+	if len(chunk.Blocks) == 0 {
+		return nil, errors.New("number of blocks is 0")
+	}
+
+	if len(chunk.Blocks) > math.MaxUint8 {
+		return nil, fmt.Errorf("number of blocks (%d) exceeds maximum allowed (%d)", len(chunk.Blocks), math.MaxUint8)
+	}
+
+	initialL2BlockNumber := chunk.Blocks[0].Header.Number.Uint64()
```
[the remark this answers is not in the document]
jonastheis: done in de09af4bc445412cf6fb4933a0213143b54b8b23

encoding/codecv7.go
```diff
@@ -0,0 +1,446 @@
...
+func (d *DACodecV7) NewDAChunk(chunk *Chunk, totalL1MessagePoppedBefore uint64) (DAChunk, error) {
+	if chunk == nil {
+		return nil, errors.New("chunk is nil")
+	}
+
+	if len(chunk.Blocks) == 0 {
+		return nil, errors.New("number of blocks is 0")
+	}
+
+	if len(chunk.Blocks) > math.MaxUint8 {
```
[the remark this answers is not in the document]
jonastheis: done in  de09af4bc445412cf6fb4933a0213143b54b8b23

encoding/codecv7_types.go
```diff
@@ -0,0 +1,499 @@
...
+	blobEnvelopeV7OffsetPayload        = 5
+)
+
+// Below is the encoding for blobPayloadV7.
+//   * Field                       Bytes     Type             Index       Comments
+//   * prevL1MessageQueueHash	   32        bytes32          0           hash of the L1 message queue at the end of previous batch
+//   * postL1MessageQueueHash      32        bytes32          32          hash of the L1 message queue at the end of this batch
+//   * initialL2BlockNumber        8         uint64           64          The initial L2 block number in this batch
+//   * numBlocks                   2         uint16           72          The number of blocks in this batch
+//   * block[0]                    52        BlockContextV2   74          The first block in this batch
```
Thegaram, line 54: Isn't type `DABlockV7` here?

encoding/codecv7_types.go
```diff
@@ -0,0 +1,499 @@
...
+	if len(b.blocks) > 0 {
+		initialL2BlockNumber = b.blocks[0].Header.Number.Uint64()
+		binary.BigEndian.PutUint64(payloadBytes[blobPayloadV7OffsetInitialL2BlockNumber:blobPayloadV7OffsetNumBlocks], initialL2BlockNumber)
+	}
+	binary.BigEndian.PutUint16(payloadBytes[blobPayloadV7OffsetNumBlocks:blobPayloadV7OffsetBlocks], uint16(len(b.blocks)))
+
+	var l1MessageIndex *uint64
+	var transactionBytes []byte
+	for i, block := range b.blocks {
+		if !block.Header.Number.IsUint64() {
```
Thegaram, line 254: We're a bit inconsistent about these sanity checks, there's a few places where we don't do them.

encoding/codecv7_types.go
```diff
@@ -0,0 +1,499 @@
...
+	return &daChunkV7{
+		daChunkV1{
+			blocks:       blocks,
+			transactions: transactions,
+		},
+	}
+}
+
+// Hash computes the hash of the DAChunk data.
+func (c *daChunkV7) Hash() (common.Hash, error) {
```
Thegaram, line 442: Should we remove this impl?

Thegaram (approved):

encoding/codecv7_types.go
jonastheis: It is used from the relayer to generate the hash of a chunk which is also a unique key in the DB: https://github.com/scroll-tech/scroll/blob/bc8f9dbc83e662aa06db161b0e4dc686a6b5818d/rollup/internal/utils/utils.go#L144. I added it later to make it compatible with how the relayer works right now. But as discussed before, I think it makes sense to clean up relayer and codec once the hardfork happened.

encoding/codecv7.go
```diff
@@ -0,0 +1,442 @@
...
+				if chunkBlocksCount > totalBlocks {
+					return false, errors.New("chunks contain more blocks than the batch")
+				}
+
+				if b.Blocks[chunkBlocksCount].Header.Hash() != block.Header.Hash() {
+					return false, errors.New("blocks in chunks do not match the blocks in the batch")
+				}
+				chunkBlocksCount++
+			}
+		}
```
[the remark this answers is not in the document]
jonastheis: Added in 01bd9b5cfff474219eb15ce363937a12847e9451

encoding/codecv7.go
```diff
@@ -0,0 +1,442 @@
...
+
+	// read the compressed flag and decompress if needed
+	compressed := rawBytes[blobEnvelopeV7OffsetCompressedFlag]
+	if compressed == 0x1 {
+		var err error
+		if payloadBytes, err = decompressV7Bytes(payloadBytes); err != nil {
+			return nil, fmt.Errorf("failed to decompress blob payload: %w", err)
+		}
+	}
+
```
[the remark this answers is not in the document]
jonastheis: good point, added a check for this in 01bd9b5cfff474219eb15ce363937a12847e9451

encoding/codecv7_types.go
jonastheis: I did some refactoring in 01bd9b5cfff474219eb15ce363937a12847e9451 and now the sanity checks should all reuse the same functions with hopefully all the sanity checks.

encoding/codecv7_types.go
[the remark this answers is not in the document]
colinlyguo: makes sense.

encoding/da.go
```diff
@@ -209,8 +261,7 @@ func convertTxDataToRLPEncoding(txData *types.TransactionData) ([]byte, error) {
 			S:          txData.S.ToInt(),
 		})
 
-	case types.L1MessageTxType: // L1MessageTxType is not supported
-	default:
+	default: // BlobTxType, SetCodeTxType, L1MessageTxType
```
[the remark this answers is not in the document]
Thegaram: True, `SetCodeTxType` will be part of the blob. Can you add this in a separate PR? @colinlyguo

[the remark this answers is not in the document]
colinlyguo: OK.

roynalnaruto (approved):

roynalnaruto (commented):
Looks good. However, I think its best to add some compatibility tests between the circuit-side and da-codec just to avoid any implementation mismatch

Thegaram (dismissed):

Thegaram (approved):

colinlyguo (approved):

PR #50, opened 2025-03-27
fix: codecv7 challenge digest

colinlyguo:
### PR title

- [x] fix: A bug fix

### Breaking change label

## Summary by CodeRabbit

- **New Features**
  - Introduced a cryptographic challenge digest into data batch creation and blob payload generation, enhancing integrity checks and data proof validation.

- **Refactor**
  - Streamlined methods and interfaces across multiple encoding versions to consistently incorporate and expose the new challenge digest for improved reliability.
  
- **Bug Fixes**
  - Updated error handling to ensure consistent return values, including the challenge digest, across various methods.

### DIFF
--- a/encoding/codecv1.go
+++ b/encoding/codecv1.go
@@ -456,3 +456,20 @@ func (d *DACodecV1) computeBatchDataHash(chunks []*Chunk, totalL1MessagePoppedBe
 	dataHash := crypto.Keccak256Hash(dataBytes)
 	return dataHash, nil
 }
+
+// ChallengeDigestFromBlobBytes calculates the challenge digest from the given blob bytes.
+func (d *DACodecV1) ChallengeDigestFromBlobBytes(blobBytes []byte) (common.Hash, error) {
+	blob, err := makeBlobCanonical(blobBytes)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to convert blobBytes to canonical form: %w", err)
+	}
+
+	c, err := kzg4844.BlobToCommitment(blob)
+	if err != nil {
+		return common.Hash{}, fmt.Errorf("failed to create blob commitment: %w", err)
+	}
+	blobVersionedHash := kzg4844.CalcBlobHashV1(sha256.New(), &c)
+
+	challengeDigest := crypto.Keccak256Hash(crypto.Keccak256(blobBytes), blobVersionedHash[:])
+	return challengeDigest, nil
+}
--- a/encoding/codecv1.go
+++ b/encoding/codecv1.go
@@ -3,7 +3,6 @@ package encoding
 import (
 	"crypto/sha256"
 	"encoding/binary"
-	"encoding/hex"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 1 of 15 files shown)


roynalnaruto (approved):

encoding/codecv1.go
```diff
@@ -456,3 +457,38 @@ func (d *DACodecV1) computeBatchDataHash(chunks []*Chunk, totalL1MessagePoppedBe
...
+		return common.Hash{}, kzg4844.Commitment{}, kzg4844.Proof{}, fmt.Errorf("failed to convert blobBytes to canonical form: %w", err)
+	}
+
+	commitment, err := kzg4844.BlobToCommitment(blob)
+	if err != nil {
+		return common.Hash{}, kzg4844.Commitment{}, kzg4844.Proof{}, fmt.Errorf("failed to create blob commitment: %w", err)
+	}
+	blobVersionedHash := kzg4844.CalcBlobHashV1(sha256.New(), &commitment)
+
+	challengeDigest := crypto.Keccak256Hash(crypto.Keccak256(blobBytes), blobVersionedHash[:])
```
jonastheis, line 474: Is this correct? I thought the `challengeDigest` would be computed differently before CodecV7

encoding/codecv7_types.go
```diff
@@ -138,7 +138,14 @@ func (b *daBatchV7) Hash() common.Hash {
 
 // BlobDataProofForPointEvaluation computes the abi-encoded blob verification data.
 func (b *daBatchV7) BlobDataProofForPointEvaluation() ([]byte, error) {
-	challengeDigest := crypto.Keccak256Hash(crypto.Keccak256(b.blobBytes), b.blobVersionedHash.Bytes())
+	if len(b.blobBytes) > maxEffectiveBlobBytes {
```
jonastheis, line 141: Could we try to reuse code here? we should try to avoid duplicate code

encoding/codecv1.go
colinlyguo: yes. It's correct. The differences are only introduced in v7.

roynalnaruto: wait why is da-codec@v1 being updated?

georgehao (approved):

encoding/codecv7.go
```diff
@@ -121,23 +122,28 @@ func (d *DACodecV7) constructBlob(batch *Batch) (*kzg4844.Blob, common.Hash, []b
...
 	c, err := kzg4844.BlobToCommitment(blob)
 	if err != nil {
-		return nil, common.Hash{}, nil, fmt.Errorf("failed to create blob commitment: %w", err)
+		return nil, common.Hash{}, nil, common.Hash{}, fmt.Errorf("failed to create blob commitment: %w", err)
 	}
 	blobVersionedHash := kzg4844.CalcBlobHashV1(sha256.New(), &c)
 
-	return blob, blobVersionedHash, blobBytes, nil
+	paddedBlobBytes := make([]byte, maxEffectiveBlobBytes)
+	copy(paddedBlobBytes, blobBytes)
```
jonastheis, line 142: Why copying here? Also wouldn't we need to call `paddedBlobBytes = makeBlobCanonical(blobBytes)` again?

Why not just use `blob` here? `blob` is padded and correct size

jonastheis (commented):
would also be good to add a check for the `challengeDigest` to the tests at least for `CodecV7`

encoding/codecv7.go
colinlyguo: `paddedBlobBytes = makeBlobCanonical(blobBytes)` this is after 31 bytes per 32 bytes conversion.

colinlyguo:
> would also be good to add a check for the `challengeDigest` to the tests at least for `CodecV7`

good suggestion. @roynalnaruto could you provide a test case here?

encoding/codecv7.go
colinlyguo: comments are added in: [5f1aec7](https://github.com/scroll-tech/da-codec/pull/50/commits/5f1aec7a8aee898843a0ca785cf8cffca6d5b097)

jonastheis (approved):

georgehao (approved):

PR #51, opened 2025-04-04
[DO NOT MERGE YET] Turn rs-zstd into official version

noel2004:
With [update in zstd decompression circuit](https://github.com/scroll-tech/rust-zstd-decompressor/pull/4), we can fully support zstd compressed data generated from official implement, with current configurations (raw literal, no magic bytes, no custom dictionary, etc). 

Since no hacking in zstd source is needed anymore once the new decompression circuit has been merged into new release of prover (maybe in the f-release?), we can turn the depending of zstd into official one safely.

NOTICE: DO NOT apply this PR before the new decompression circuit has been applied.


libzstd/encoder/Cargo.toml
```diff
@@ -7,4 +7,4 @@ edition = "2021"
 [lib]
 
 [dependencies]
-zstd = { git = "https://github.com/scroll-tech/zstd-rs", branch = "hack/mul-block", features = ["experimental"]}
\ No newline at end of file
+zstd = { version = "0.13", features = ["experimental"]}
```
colinlyguo, line 10: I'm concerned that this will affect the compression of codec version < 7, thus affecting the rollup verifier's batch hash calculation. @jonastheis, would the current L1 DA syncer replace constructing local blocks and compress them to a batch? e.g., replace by getting blob raw data and decompress it, then verify against local blocks.)

noel2004: Yes. I think the later is better designation and we should finally adopt to that. Making compression more than once and expect the result is identify seems a bit danger because the result would be affected by many factors: any update of minor version, configuration etc.

colinlyguo: Quite agree with the claim.

encoding/da.go
```diff
@@ -509,6 +509,15 @@ func checkCompressedDataCompatibilityV7(data []byte) error {
 	return nil
 }
 
+// Fast testing if the compressed data (v8) is compatible with our circuit
+// (require specified frame header and each block is compressed)
+func checkCompressedDataCompatibilityV8(data []byte) error {
```
colinlyguo, line 514: Currently changing this function, only keeping a very simple check (other checks are failing in unit tests). cc @noel2004 to check if it's safe to do this.

libzstd/encoder-standard/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+pub fn init_zstd_encoder(target_block_size: u32) -> Encoder<'static, Vec<u8>> {
+    let mut encoder = Encoder::new(Vec::new(), 0).expect("infallible");
+
+    // disable compression of literals, i.e. literals will be raw bytes.
+    encoder
+        .set_parameter(CParameter::LiteralCompressionMode(ParamSwitch::Disable))
+        .expect("infallible");
+    // with a hack in zstd we can set window log <= 17 with single segment kept
+    encoder
+        .set_parameter(CParameter::WindowLog(17))
```
noel2004, line 32: For new encoder, increase the windowLog from 17 -> 22 (128K -> 4M), it would help to improve the compressing ratio and enable us a signle frame scheme for any source data less than 4MB (should be enough in quite a long peroid)

colinlyguo: Changed in this commit: [b64a7d3](https://github.com/scroll-tech/da-codec/pull/51/commits/b64a7d3cfc1046e8ae4fbd193b87fa2ba8348c84) and regenerated the `.a` files.

PR #52, opened 2025-05-12
fix: check v7 batch compressed data compatibility

colinlyguo:

For v7 batches, uncompressed cases are resolved. This PR removes the uncompressed case checks and only keeps sanity checks. If sanity check fails, rollup-relayer should halt and wait for manual fix.

### PR title

- [ ] fix: A bug fix

### Breaking change label

### DIFF
--- a/encoding/codecv7.go
+++ b/encoding/codecv7.go
@@ -231,7 +231,7 @@ func (d *DACodecV7) checkCompressedDataCompatibility(payloadBytes []byte) ([]byt
 		return nil, false, fmt.Errorf("failed to compress blob payload: %w", err)
 	}
 
-	if err = checkCompressedDataCompatibility(compressedPayloadBytes); err != nil {
+	if err = checkCompressedDataCompatibilityV7(compressedPayloadBytes); err != nil {
 		log.Warn("Compressed data compatibility check failed", "err", err, "payloadBytes", hex.EncodeToString(payloadBytes), "compressedPayloadBytes", hex.EncodeToString(compressedPayloadBytes))
 		return nil, false, nil
 	}
--- a/encoding/codecv7_test.go
+++ b/encoding/codecv7_test.go
@@ -384,18 +384,6 @@ func TestCodecV7BatchStandardTestCasesEnableCompression(t *testing.T) {
 		})
 	}
 
-	repeat := func(element byte, count int) string {
-		result := make([]byte, 0, count)
-		for i := 0; i < count; i++ {
-			result = append(result, element)
-		}
-		return "0x" + common.Bytes2Hex(result)
-	}
-
-	// Taking into consideration compression, we allow up to 5x of max blob bytes minus 5 byte for the blob envelope header.
-	// We subtract 74 bytes for the blobPayloadV7 metadata.
-	//compressableAvailableBytes := maxEffectiveBlobBytes*5 - 5 - blobPayloadV7MinEncodedLength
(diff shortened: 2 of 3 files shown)


encoding/codecv7_test.go
```diff
@@ -384,18 +384,6 @@ func TestCodecV7BatchStandardTestCasesEnableCompression(t *testing.T) {
 		})
 	}
 
-	repeat := func(element byte, count int) string {
```
noel2004, line 387: Why we should erase this testing case?

noel2004 (dismissed):

encoding/codecv7_test.go
colinlyguo: because it's an unable to compress test case. or we can add it back and change the versioned hash. wdyt?

encoding/codecv7.go
```diff
@@ -267,7 +276,7 @@ func (d *DACodecV7) CheckBatchCompressedDataCompatibility(b *Batch) (bool, error
 		return false, fmt.Errorf("failed to construct blob payload: %w", err)
 	}
 
-	_, compatible, err := d.checkCompressedDataCompatibility(payloadBytes)
+	_, compatible, err := d.checkCompressedDataCompatibility(payloadBytes, false /* checkLength */)
```
noel2004, line 279: Maybe we should add some comment to illustrate that why we should skip length checking here.

noel2004 (dismissed):

encoding/da.go
```diff
@@ -465,6 +465,52 @@ func checkCompressedDataCompatibility(data []byte) error {
...
+	for len(data) > 3 && !isLast {
+		isLast = (data[0] & 1) == 1
+		blkSize := (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
+		if len(data) < 3+int(blkSize) {
+			return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
+		}
+		data = data[3+blkSize:]
+	}
+
+	// Should we return invalid if isLast is still false?
```
jonastheis, line 506: remove this comment

encoding/da.go
```diff
@@ -465,6 +465,52 @@ func checkCompressedDataCompatibility(data []byte) error {
 	return nil
 }
 
+// Fast testing if the compressed data (v7) is compatible with our circuit
+// (require specified frame header and each block is compressed)
+func checkCompressedDataCompatibilityV7(data []byte) error {
```
jonastheis, line 470: what exactly is the difference between this method and `checkCompressedDataCompatibility`?

colinlyguo: the only difference is this check is removed:
```
		if blkType != 2 {
			return fmt.Errorf("unexpected blk type {%d}, size {%d}, last {%t}", blkType, blkSize, isLast)
		}
```
want to keep the behavior consistent for previous codec versions, so adding a new function, but it's open to change (might also affect some unit test results of previous codec versions).

encoding/da.go
colinlyguo: removed in [feb63c9](https://github.com/scroll-tech/da-codec/pull/52/commits/feb63c9ee30a92f5f539603fb5c0d275ea290cb3).

encoding/codecv7.go
colinlyguo: makes sense. added in [feb63c9](https://github.com/scroll-tech/da-codec/pull/52/commits/feb63c9ee30a92f5f539603fb5c0d275ea290cb3).

jonastheis (approved):

noel2004 (approved):

PR #64, opened 2025-10-03
fix: consider RLE blocks in zstd compatibility check

noel2004:
Fix an issue that the new `checkCompressedDataCompatibilityV7` do not consider RLE type block, which would be possible when we have switched to official zstd for data compression

After this change, `checkCompressedDataCompatibilityV7` will serve as a sanity check, it will only return error if there is some issue with the compressed data. Examples include: Data is corrupted, data is truncated, compressed data uses unsupported features (e.g. dictionary). If the compression module is configured correctly, `checkCompressedDataCompatibilityV7` will never fail.

See the spec for more details: https://datatracker.ietf.org/doc/html/rfc8878

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of compressed data blocks to reduce decoding failures.
  * Correctly interprets block types, enforcing proper size for RLE blocks and rejecting reserved types with clear errors.
  * Provides more accurate data length checks and error reporting, improving reliability when processing V7-compressed inputs.

### DIFF
--- a/encoding/da.go
+++ b/encoding/da.go
@@ -497,13 +497,13 @@ func checkCompressedDataCompatibilityV7(data []byte) error {
 		isLast = (data[0] & 1) == 1
 		blkType := (data[0] >> 1) & 3
 		var blkSize uint
-		if blkType == 1 {// RLE Block
+		if blkType == 1 { // RLE Block
 			blkSize = 1
 		} else {
 			if blkType == 3 {
 				return fmt.Errorf("encounter reserved block type at %v", data)
 			}
-			blkSize = (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3	
+			blkSize = (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
 		}
 		if len(data) < 3+int(blkSize) {
 			return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
--- a/encoding/da.go
+++ b/encoding/da.go
@@ -495,7 +495,16 @@ func checkCompressedDataCompatibilityV7(data []byte) error {
 	// scan each block until done
 	for len(data) > 3 && !isLast {
 		isLast = (data[0] & 1) == 1
-		blkSize := (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
+		blkType := (data[0] >> 1) & 3
+		var blkSize uint
+		if blkType == 1 {// RLE Block
+			blkSize = 1
+		} else {
+			if blkType == 3 {
+				return fmt.Errorf("encounter reserved block type at %v", data)
+			}
+			blkSize = (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3	
+		}
 		if len(data) < 3+int(blkSize) {
 			return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
 		}


encoding/da.go
```diff
@@ -495,7 +495,16 @@ func checkCompressedDataCompatibilityV7(data []byte) error {
 	// scan each block until done
 	for len(data) > 3 && !isLast {
 		isLast = (data[0] & 1) == 1
-		blkSize := (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
+		blkType := (data[0] >> 1) & 3
+		var blkSize uint
+		if blkType == 1 { // RLE Block
+			blkSize = 1
```
jonastheis, line 501: Why set this to 1 here?

noel2004: According to RFC8878, `blkType` is 1 indicating this is a RLE block, and such a block [has a fixed 1-byte size](https://datatracker.ietf.org/doc/html/rfc8878#name-block_type):

> RLE_Block:
> This is a single byte, repeated Block_Size times. Block_Content consists of a single byte. On the decompression side, this byte must be repeated Block_Size times.

jonastheis (approved):
I think this change should be relatively low risk as we decode batches before submitting them in the rollup relayer anyway.

As discussed in Slack, we should probably replace this check long-term with a decompression with the underlying library to avoid such compatibility issues and be better maintainable.

encoding/da.go
```diff
@@ -495,7 +495,16 @@ func checkCompressedDataCompatibilityV7(data []byte) error {
...
+		} else {
+			if blkType == 3 {
+				return fmt.Errorf("encounter reserved block type at %v", data)
+			}
+			blkSize = (uint(data[2])*65536 + uint(data[1])*256 + uint(data[0])) >> 3
+		}
 		if len(data) < 3+int(blkSize) {
 			return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
```
Thegaram, line 509: Would this happen in practice?

encoding/da.go
Thegaram, line 468: Seems like this is not true anymore, we support all block types (raw, RLE, compressed), right?

So what does this function check exactly?

Thegaram, line 483: The [spec](https://datatracker.ietf.org/doc/html/rfc8878#name-frame_header_descriptor) says
> When Frame_Content_Size_Flag is 0, FCS_Field_Size depends on Single_Segment_Flag: if Single_Segment_Flag is set, FCS_Field_Size is 1. Otherwise, FCS_Field_Size is 0

Is `Single_Segment_Flag` never set in our case?

Thegaram (approved):

encoding/da.go
noel2004: Yes. It is found by @jonastheis while encoding the on-chain block (cause some rather small batch).

Consider many 0 bytes is filled in calldata of a tx, it would be highly possible for zstd had generated a RLE block.

encoding/da.go
noel2004: There is still reserved value (3) and we need to ensure a  block type can not be that (broken data)

noel2004: It should be ALWAYS set in our case. That is why we [have expected the leftist 5 bit is 32](https://github.com/scroll-tech/da-codec/blob/4e9428c3cd82a74b8811e32e9b773e0edce77e77/encoding/da.go#L476) in frame header.

encoding/da.go
Thegaram: I see what you mean. So this error happened before because of an RLE block:

```
return fmt.Errorf("wrong data len {%d}, expect min {%d}", len(data), 3+blkSize)
```

Now that we added a special case for RLE blocks, this error should normall not happen.

encoding/da.go
Thegaram: But that would not happen if we use a correct zstd encoder implementation. So this is more like a sanity check now.

PR #66, opened 2025-11-14
add galileo CodecV9

georgehao:

This PR adds CodecV9 for Galileo. This codec is identical with CodecV8, except that it uses version `9` in both the batch header and blob envelope.

### PR title

- [X] feat: A new feature

### Breaking change label

- [X] No, this PR is not a breaking change
- [ ] Yes

## Summary by CodeRabbit

* **New Features**
  * Adds Codec V9 support (selected when Galileo hardfork is active), including V9-aware compression, blob construction, and L1 commit size estimation for batches/chunks.

* **Tests**
  * Adds tests verifying Codec V9 selection and mapping, plus an error case for a future version.

* **Chores**
  * Bumps Go toolchain to 1.22 and refreshes cryptography/KZG-related dependencies; updates testify to v1.10.0.


encoding/codecv9.go
```diff
@@ -0,0 +1,14 @@
+package encoding
+
+type DACodecV9 struct {
```
Thegaram, line 29: Please confirm that this is true: We do not need any of the added complexities (duplicated methods) that we needed for V8, because we do not change the internal compression behavior in V9.

Thegaram (dismissed):

encoding/codecv9.go
Thegaram, line 1: Remember that we need to bump `da-codec` version in both `rollup-relayer` and `l2geth` (I think previously we forgot about l2geth).

encoding/codecv9.go
jonastheis: I think so. Just like CodecV6

jonastheis (dismissed):

encoding/codecv9.go
yiweichi: blob-uploader depends on da-code as well, Do we already have plan to update blob-uploader [here](https://github.com/scroll-tech/scroll/blob/develop/rollup/internal/controller/blob_uploader/blob_uploader.go#L170) to support codecv9?

yiweichi: Nevermind, it's already covered on pr https://github.com/scroll-tech/scroll/pull/1752/files#diff-516f715f986f16ab3c638b2e3652d9bf4af757c33fd880a0964dea26c2d48f2dR1053.

georgehao: added in this pr https://github.com/scroll-tech/scroll/pull/1752

PR #67, opened 2025-11-14
feat: consider RLE blocks in compatibility check starting from V9

Thegaram:

Enable https://github.com/scroll-tech/da-codec/pull/64 starting from codecV9.

See also [discussion](https://scrollco.slack.com/archives/C02PKAQ2430/p1762959428487289?thread_ts=1758877774.049689&cid=C02PKAQ2430).

### PR title

- [X] feat: A new feature

### Breaking change label

- [X] No, this PR is not a breaking change
- [ ] Yes

### DIFF
--- a/encoding/codecv9.go
+++ b/encoding/codecv9.go
@@ -12,8 +12,8 @@ import (
 	"github.com/scroll-tech/go-ethereum/log"
 )
 
-// DACodecV9 updates the implementation of checkCompressedDataCompatibility to use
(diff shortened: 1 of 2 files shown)


jonastheis (approved):
Looks good to me (checked with a diff to `CodecV8`). 

Just to clarify:
Once Galileo activates we start creating batches with `CodecV9` and potentially with RLE blocks. This is fine as l2geth will then (after Galileo) recreate the same batches with RLE blocks. Compression is unmodified and should work in either case.

georgehao (approved):

